PR #14, opened 2025-06-04
chore: add migrate step to local dev setup in readme

donkeysharp:
I was following the steps and unfortunately I wasn't able to run by 100% following the README instructions. The db:migrate step was missing


hjball:
Good spot

PR #18, opened 2025-06-04
feat(auth): added discord & github authentication methods

LovelessCodes:
I've added a way to tell if Google, Discord or Github auth environment variables have been set, and only display the "Continue with" buttons, if they're set.


hjball (approved):
Nice work, tested locally lgtm

PR #28, opened 2025-06-05
feat: add S3 avatar upload during user creation from social provider

### LINKED ISSUE #23: User avatar images set by Google OAuth do not load
## 🐛 Bug Report

**Describe the bug**
Creating an account with Google OAuth sets the user avatar to the Google account image which does not load correctly.

**To Reproduce**

1. Create new user with Google OAuth
2. Go to /settings
3. See broken user avatar image

**Expected behavior**

If the user has a URL which doesn't match the S3 format for avatars we should use the full imageUrl

LovelessCodes:
Should fix #23

### DIFF
--- a/packages/auth/src/auth.ts
+++ b/packages/auth/src/auth.ts
@@ -90,8 +90,6 @@ export const initAuth = (db: dbClient) => {
                   },
                 });
 
-                console.log(user.image);
-
                 const allowedFileExtensions = ["jpg", "jpeg", "png", "webp"];
 
                 const fileExtension = user.image.split('.').pop()?.split('?')[0] || 'jpg';
--- a/packages/auth/src/auth.ts
+++ b/packages/auth/src/auth.ts
@@ -4,6 +4,7 @@ import { createAuthMiddleware } from "better-auth/api";
 import { apiKey } from "better-auth/plugins";
 import { magicLink } from "better-auth/plugins/magic-link";
 import { env } from "next-runtime-env";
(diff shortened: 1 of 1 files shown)


packages/auth/src/auth.ts
```diff
@@ -65,6 +74,47 @@ export const initAuth = (db: dbClient) => {
...
+                  credentials: {
+                    accessKeyId: env("S3_ACCESS_KEY_ID") ?? "",
+                    secretAccessKey: env("S3_SECRET_ACCESS_KEY") ?? "",
+                  },
+                });
+
+                const allowedFileExtensions = ["jpg", "jpeg", "png", "webp"];
+
+                const fileExtension = user.image.split('.').pop()?.split('?')[0] || 'jpg';
+                const key = `${Date.now()}-${Math.random().toString(36).substring(2, 15)}.${!allowedFileExtensions.includes(fileExtension) ? 'jpg' : fileExtension}`;
```
hjball, line 96: Can we update this to follow the same key format as uploads from the settings page:

```
[userId]/avatar[extension]
aa6af775-46c1-46a3-a815-c87d5cd8822d/avatar.jpg
```

hjball (commented):
Good stuff, just one comment

hjball (approved):

packages/auth/src/auth.ts
hjball: Nice one, thanks

PR #31, opened 2025-06-05
fix: add cascade delete for card-related foreign keys and invalidate …

### LINKED ISSUE #30: Attempting to delete workspace with a card which has a label assigned, it fails
## 🐛 Bug Report

**Describe the bug**
When a label is attached to a card, it creates a card_activity entry, and the card_activity entry can't be deleted when it has a label id connected.
This creates a foreign key constraint.

**To Reproduce**

1. Create a new workspace
2. Create a board on said workspace
3. Create a list on said board
4. Create a card on said list
5. Create a label on said card
6. Attach the label on said card
7. Go to settings
8. Click Delete workspace
9. Attempt to Delete workspace
10. See that it fails

**Expected behavior**
Should delete said workspace.

LovelessCodes:
…workspace cache after deletion

Closes #30


hjball (commented):
Don't forget to generate the new migration:

```
cd packages/db && pnpm drizzle-kit generate --name="AddCascadeDeleteToCardRelations"
```

```
pnpm with-env npx tsx ./src/migrate.ts
```

PR #32, opened 2025-06-05
fix: handle external URLs in avatar rendering (#23)

### LINKED ISSUE #23: User avatar images set by Google OAuth do not load

lvuur-godaddy:
Fixes #23 

## Description
Fixed a bug where Google OAuth avatars were not displaying correctly in the settings page. When a user creates an account with Google OAuth, the avatar URL is a full Google URL, but our getAvatarUrl function was prepending our storage bucket path, resulting in broken image links.
   
## Changes Made

    - Updated `getAvatarUrl` function in `apps/web/src/utils/helpers.ts` to detect full URLs
    - Added conditional logic to return URLs as-is when they start with `http://` or `https://`
    - Renamed parameter from `key` to `imageOrKey` for better clarity
    - Maintains backward compatibility with existing S3 key usage


LovelessCodes:
> Good start, I think you'll need to update the remotePatterns in the next config to allow lh3.googleusercontent.com domains
> 
> [nextjs.org/docs/app/getting-started/images#remote-images](https://nextjs.org/docs/app/getting-started/images#remote-images)

Is Google the only provider you'll support, going forward?
If it is, it's fine doing it this way, if not maybe consider #28 instead.

hjball:
> > Good start, I think you'll need to update the remotePatterns in the next config to allow lh3.googleusercontent.com domains
> > [nextjs.org/docs/app/getting-started/images#remote-images](https://nextjs.org/docs/app/getting-started/images#remote-images)
> 
> Is Google the only provider you'll support, going forward? If it is, it's fine doing it this way, if not maybe consider #28 instead.

For the time being, yes, but this is a quick win for users who have already signed up. We should be uploading all provider avatars to s3 going forward (like you've proposed in #28).

hjball:
> Good start, I think you'll need to update the remotePatterns in the next config to allow lh3.googleusercontent.com domains
> 
> https://nextjs.org/docs/app/getting-started/images#remote-images

The error is clogging up my logs so I've pushed up the changes @lvuur-godaddy

PR #34, opened 2025-06-05
feat: add support for 15 new OAuth providers including Apple, Microsoft, and TikTok

LovelessCodes:
…ft, and TikTok


.env.example
```diff
@@ -25,4 +25,35 @@ DISCORD_CLIENT_ID=
 DISCORD_CLIENT_SECRET=
 GITHUB_CLIENT_ID=
 GITHUB_CLIENT_SECRET=
+GITLAB_CLIENT_ID=
```
hjball, line 28: This feels like too env heavy which could potentially be off putting for new people to the repo

I'd prefer to see the full list of supported OAuth providers in the docs, but we can add that at a later date

README.md
```diff
@@ -92,6 +92,37 @@ pnpm dev
 | `DISCORD_CLIENT_SECRET`          | Discord OAuth client secret   | For Discord login | `xxx`                                         |
 | `GITHUB_CLIENT_ID`               | GitHub OAuth client ID        | For GitHub login | `xxx`                                         |
 | `GITHUB_CLIENT_SECRET`           | GitHub OAuth client secret    | For GitHub login | `xxx`                                         |
+| `GITLAB_CLIENT_ID`               | GitLab OAuth client ID        | For GitLab login | `xxx`                                         |
```
hjball, line 95: Same here

.env.example
LovelessCodes: Done. 😄

hjball (approved):
good stuff

.env.example
hjball: Cheers

hjball: Discord link if you're not already in it -> https://discord.gg/e6ejRb6CmT

PR #38, opened 2025-06-05
refactor: update board slug prefix to use dynamic base URL from env

### LINKED ISSUE #37: Updating board URL, it uses kan.bn by default
## 🐛 Bug Report

**Describe the bug**
When attempting to edit the board url, it uses kan.bn instead of the base url by default.

**To Reproduce**

1. Clone the repository
2. Set environment variables
3. Run Kan locally
4. Go to a board
5. Click on the three dots in the top right corner
6. Click Edit board URL
7. See that it uses kan.bn by default

**Expected behavior**
To use the set base url, so that people can copy it.

LovelessCodes:
closes #37

### DIFF
--- a/packages/api/src/routers/board.ts
+++ b/packages/api/src/routers/board.ts
@@ -365,12 +365,12 @@ export const boardRouter = createTRPCRouter({
       })
       .input(
         z.object({
-          boardPublicId: z.string().min(12),
           boardSlug: z
             .string()
             .min(3)
             .max(24)
             .regex(/^(?![-]+$)[a-zA-Z0-9-]+$/),
+          boardPublicId: z.string().min(12),
         }),
       )
       .output(
@@ -379,19 +379,13 @@ export const boardRouter = createTRPCRouter({
         }),
       )
       .query(async ({ ctx, input }) => {
-        const slug = input.boardSlug.toLowerCase();
-        const workspace = await boardRepo.getWorkspaceAndBoardIdByBoardPublicId(ctx.db, input.boardPublicId);
-        if (!workspace)
-          throw new TRPCError({
-            message: `Board with public ID ${input.boardPublicId} not found`,
-            code: "NOT_FOUND",
-          });
-        const isSlugUnique = await boardRepo.isSlugUnique(ctx.db, {
-          slug,
-          workspaceId: workspace.workspaceId,
-        })
+        const isBoardSlugAvailable = await boardRepo.isBoardSlugAvailable(
+          ctx.db,
+          input.boardSlug,
+          input.boardPublicId,
+        );
         return {
-          isReserved: !isSlugUnique,
+          isReserved: !isBoardSlugAvailable,
         };
       }),
 });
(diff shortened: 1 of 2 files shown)


packages/db/src/repository/boardSlug.repo.ts
```diff
@@ -0,0 +1,13 @@
+import { eq } from "drizzle-orm";
+
+import type { dbClient } from "@kan/db/client";
+import { boards } from "@kan/db/schema";
+
+export const getBoardSlug = (db: dbClient, slug: string) => {
```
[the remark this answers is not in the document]
hjball: Also please could you move this to the board repo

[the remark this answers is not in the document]
LovelessCodes: Should've been done now.

packages/api/src/routers/board.ts
```diff
@@ -352,4 +352,37 @@ export const boardRouter = createTRPCRouter({
 
       return { success: true };
     }),
+    checkSlugAvailability: publicProcedure
+      .meta({
+        openapi: {
+          summary: "Check if a board slug is available",
+          method: "GET",
+          path: "/boards/check-slug-availability",
```
hjball, line 360: I think it makes more sense to pass in the boardPublicId

```suggestion
          path: "/boards/{boardPublicId}/check-slug-availability",
```

packages/api/src/routers/board.ts
```diff
@@ -352,4 +352,37 @@ export const boardRouter = createTRPCRouter({
...
+        },
+      })
+      .input(
+        z.object({
+          boardSlug: z
+            .string()
+            .min(3)
+            .max(24)
+            .regex(/^(?![-]+$)[a-zA-Z0-9-]+$/),
+          workspaceSlug: z.string(),
```
hjball, line 373:
```suggestion

```

packages/db/src/repository/board.repo.ts
```diff
@@ -477,3 +477,27 @@ export const getWorkspaceAndBoardIdByBoardPublicId = async (
...
+      }
+    },
+    where: eq(boards.slug, boardSlug),
+  });
+
+  return {
+    id: result?.id,
+    workspaceSlug: result?.workspace.slug,
+  };
+};
```
hjball:
```suggestion
export const isBoardSlugAvailable = async (
  db: dbClient,
  boardSlug: string,
  boardPublicId: string,
) => {
  const result = await db.query.boards.findFirst({
    columns: {
      id: true,
    },
    where: and(
      eq(board.publicId, boardPublicId)
      eq(board.slug, boardSlug),
      isNull(board.deletedAt),
    ),
  });

  return result === undefined;
};
```

LovelessCodes:
I chose to use some existing functions in the repos.

hjball:
> I chose to use some existing functions in the repos.

Ah right, I think this warrants it's own repo func. This will currently return the first board that matches that slug (potentially thousands for a common slug like Roadmap)

packages/db/src/repository/board.repo.ts
LovelessCodes: This code is kinda weird - because it tries to look for a board with the same publicId AND same slug.
The weirdness of it, is when you try to look for the same publicId.
It should look for the same workspaceId though.

hjball: You're right, I had completely overlooked that (was obviously tired lol)

hjball (approved):
Good stuff

PR #42, opened 2025-06-05
feat: add template selection when creating new boards with predefined…

LovelessCodes:
… lists


hjball:
This is a great start for simple board templates. I want to make some small UI/UX tweaks before merging to keep it inline with the platform, but good work.

hjball:
Sorry it took me a while to get round to this one, but it's finally ready to ship @LovelessCodes

PR #43, opened 2025-06-05
Feat: Markdown editor for description on cards

### LINKED ISSUE #22: Markdown editor for card descriptions
## ✨ Feature Request

**Is your feature request related to a problem? Please describe.**

The current input field for card descriptions is very limited in functionality and buggy. We should support basic markdown editing instead of just a text field. 

**Describe the solution you'd like**
- Markdown editor with basic functionality e.g. headings, links, code blocks etc
- Should use a well maintained and flexible open source package e.g https://github.com/mdx-editor/editor
- Should still save automatically when a user clicks outside of the component
- Maintains the current styling

LovelessCodes:
closes #22


hjball:
Keen to check this one out @LovelessCodes

hjball:
Doesn't seem to be working for me locally, but I'll have a proper look this evening @LovelessCodes

hjball:
Keen to get this one out because it will fix a number of issues and vastly improve UX

hjball:
I had a good play around with this, but I couldn't quickly get UI/UX experience needed to fit the project with this package. I then tried https://gravity-ui.com/libraries/markdown-editor which has a much closer experience out-the-box, however, I ran into some compilation errors and have put it on hold.

I think we have three options:
1. Continue exploring https://gravity-ui.com/libraries/markdown-editor as a potential out-the-box option
2. Invest some time into configuring MDXEditor to be closer to the experience of the Gravity UI editor
3. Explore other options, potentially something like tip tap [https://github.com/ueberdosis/tiptap](https://github.com/ueberdosis/tiptap)

I think in the meantime we should fix bugs with saving the current content editable and make links clickable: https://github.com/nfrasser/linkifyjs

LovelessCodes:
The problem I see with Gravity UI markdown editor, is that if you choose to go that route, you're locking yourself into their design principles.
MDXEditor is pretty difficult to get good DX with, and they also provide their own set of components which also sets a limit on how much you can control.
Tiptap is a really good contender, and it's something I've worked with before - it does require some set up time, but once it's implemented, it'll be a good stepping stone.

hjball:
This is really great work @LovelessCodes 

I think with a bit of tweaking tiptap could work for us as a longer term solution. Ideally, I'd like to do away with the menu items and have everything driven by the `/` command opening up a dropdown with the available options. That would reduce clutter in the UI and keep the focus in the editor. Is this something you think we'd be able to achieve without too much effort using tiptap?

LovelessCodes:
Yeah, that sounds fine - but shouldn't there be some kind of tooltip to let people know they can do slash-commands?

hjball:
Just pushed a small fix to stop the menu items from taking focus and submitting

hjball:
> Yeah, that sounds fine - but shouldn't there be some kind of tooltip to let people know they can do slash-commands?

Maybe for the initial release we could just change the placeholder text to  "Add a description (press / to use markdown)" or something like that?

LovelessCodes:
TODO:
- [x] ~~Migrate to TipTap editor~~
- [x] ~~Migrate from traditional menu to bubble menu~~
- [x] ~~Add slash commands~~
- [ ] Add some kind of guide on how to use the description editor
    - This should include keybinds and markdown usage

PR #49, opened 2025-06-06
fix: prevent list buttons from stacking

MathieuAndrade:
Improve list header style

Before: 

image

After: 

image

### DIFF
--- a/apps/web/src/views/board/components/List.tsx
+++ b/apps/web/src/views/board/components/List.tsx
@@ -1,17 +1,16 @@
-import { type ReactNode } from "react";
-import {
-  HiOutlinePlusSmall,
-  HiEllipsisHorizontal,
-  HiOutlineTrash,
-  HiOutlineSquaresPlus,
-} from "react-icons/hi2";
+import type { ReactNode } from "react";
 import { Draggable } from "react-beautiful-dnd";
 import { useForm } from "react-hook-form";
-
-import { api } from "~/utils/api";
-import { useModal } from "~/providers/modal";
+import {
+  HiEllipsisHorizontal,
+  HiOutlinePlusSmall,
+  HiOutlineSquaresPlus,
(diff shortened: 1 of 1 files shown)


apps/web/src/views/board/components/List.tsx
```diff
@@ -80,20 +80,20 @@ export default function List({
           {...provided.dragHandleProps}
           className="dark-text-dark-1000 mr-5 h-fit min-w-[18rem] max-w-[18rem] rounded-md border border-light-400 bg-light-300 py-2 pl-2 pr-1 text-neutral-900 dark:border-dark-300 dark:bg-dark-100"
         >
-          <div className="flex justify-between">
+          <div className="mb-4 flex justify-between">
             <form
               onSubmit={handleSubmit(onSubmit)}
-              className="focus-visible:outline-none"
+              className="w-2/3 focus-visible:outline-none"
```
hjball, line 86: Ideally we want the width of the form/input to take up all of the remaining space. I assume if you set it to w-full or w-auto you're still seeing the same issue?

Also, what browser are you seeing this on?

MathieuAndrade: Hello ! 

Yea ideally the form has to take all remaining space, i agree. 
This is seem to be the input that take more space that he deserve.

If i put w-auto i effectively has the same issue -> 

image

I use Zen Browser, so i checked other browsers and it seem to be a Mozilla like browser issue !

On chrome 
image

On Edge 
image

On Firefox
image

On Floorp (Mozilla browser)
image

On Zen 
image

MathieuAndrade: Tricky but, if i put a min-w-0 on the form instead of w-2/3, it works like a charm on both browsers type.

So eventually we could use this solution instead of fixing the width ?

PR #50, opened 2025-06-07
refactor: replace js-cookie auth with new auth client and remove unused packages from web package

LovelessCodes:
jwt-decode is not used anymore, neither is js-cookie when we have better-auth client


hjball (approved):
Good stuff

PR #75, opened 2025-06-11
Feat: Reduce Docker Image Size

LovelessCodes:
What I've done:
Added an entrypoint.sh, so that the db gets migrated automatically.
Only allow db and web packages to be transported to the final docker image.

Old size: 2.8GB
New size: 1.4GB


hjball (approved):
Nice work @LovelessCodes

PR #100, opened 2025-07-05
feat: add S3_FORCE_PATH_STYLE configuration for MinIO compatibility

ReesMorris:
Adds `S3_FORCE_PATH_STYLE` environment variable to support MinIO and other S3-compatible services that require path-style URLs.

When set to `true`, uses path-style URLs (`domain.com/bucket/key`) instead of virtual hosted-style URLs (`bucket.domain.com/key`).

Also fixes ACL consistency between auth flow and upload API.


hjball (approved):
Nice work @ReesMorris

PR #115, opened 2025-07-14
fix: auth /social-providers endpoint ids

connorjan:
The `/social-providers` endpoint should return a list of the provider ids, not their names. This was found when trying to use the microsoft endpoint and the endpoint was returning `"microsoft entraid"`, when the `AuthForm` expects it to just be `"microsoft"`. This was resulting in the `AuthForm` thinking there was a provider that was supposed to be used, but the button wasn't actually being created because the name was not matching the key.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,7 @@
       "type": "node-terminal",
       "request": "launch",
       "command": "pnpm dev",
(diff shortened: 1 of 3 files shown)


hjball:
Good spot @connorjan 

Agree we could definitely do more to handle multiple OAuth providers in the UI, but I think it's lower priority compared to some of the bugfixes/features we want to ship right now!

PR #119, opened 2025-07-17
fix: creating a new label when adding a new card resets all fields

### LINKED ISSUE #25: Creating a new label when adding a new card resets all fields
## 🐛 Bug Report

**Describe the bug**
When creating a new card if you attempt to create a new label from the label dropdown all fields are reset.

**To Reproduce**

1. Create new card
2. Fill out title and description
3. Click on "Create new label" in the label dropdown
4. Add new label
5. See not returned back to card modal

**Expected behaviour**

I should be returned to the new card form with its state persisted (and the new label should be added)

Rythm18:
Fixes #25 

Enhanced the modal provider to support stacking multiple modals:
```typescript
// before - Single modal state (overwrites previous)
const [modalContentType, setModalContentType] = useState("");

// after - Modal stack (preserves previous modals)
const [modalStack, setModalStack] = useState<ModalState[]>([]);
```

**Flow:**
- `openModal("NEW_CARD")` → Stack: `["NEW_CARD"]`
- `openModal("NEW_LABEL")` → Stack: `["NEW_CARD", "NEW_LABEL"]`  
- `closeModal()` → Stack: `["NEW_CARD"]` ← **Returns to previous modal!**

There was mouting issue which on opening a nested modal, the previous modal was getting unmounted that was causing the issue that card data not being persist. So, implemented modal rendering that keeps modal components mounted if needed:

```typescript
{Object.entries(modalConfig).map(([modalType, config]) => {
  if (!config.keepMounted) return null;
  return (
    <div style={{ display: modalType === modalContentType ? 'block' : 'none' }}>
      {config.component} {/* stays mounted, preserves state */}
    </div>
  );
})}
```

Now since we are preserving the mounting of card component if we open label modal, this was leading to an issue that the card data was getting submitted because <CheckboxDropdown> is inside the form element and the button type was not defined so the default type was submit. So to prevent that:

```typescript
// before - default type="submit" caused form submission
<button onClick={() => handleCreate()}>

// after - explicit type="button" prevents form submission  
<button type="button" onClick={(e) => { e.preventDefault(); handleCreate(); }}>
```


hjball:
This is looking really nice @Rythm18 

Couple small suggestions:
- Closing the modal seems to cut the transition off short (it doesn't fade out smoothly anymore)
- For optimal UX we should automatically select the label when added

Other than that I think we're good to merge

Rythm18:
@hjball let me take a look again for the transition thing, for the second point, i guess there's a separate issue created and someone was assigned to it, so i didnt worked on it.

hjball:
You're right, don't worry about the second point for now @Rythm18 

Good to merge once the transition issue is fixed!

Rythm18:
@hjball sorry for reaching out late, i am just trying out a whole new approach since this modal stacking appraoch has broke the animation, will update the pr asap

Rythm18:
@hjball i have updated the changes, please review it!

hjball (approved):
Nice work @Rythm18 

Let's get this merged

PR #130, opened 2025-07-28
fix: disable create list and label button when input is empty

Rythm18:
Currenty when creating a new list with empty name throws error and same goes with creating new label.
The buttons will be disabled until there's some input.


hjball (approved):
Good spot

PR #138, opened 2025-08-07
feat: add pglite support

CyberL1:
Added pglite support so an external database is no longer required

### DIFF
--- a/packages/db/src/client.ts
+++ b/packages/db/src/client.ts
@@ -1,8 +1,12 @@
 import type { NodePgDatabase } from "drizzle-orm/node-postgres";
+import { drizzle as drizzlePgLite } from "drizzle-orm/pglite"; 
+import { uuid_ossp } from "@electric-sql/pglite/contrib/uuid_ossp";
 import { drizzle as drizzlePg } from "drizzle-orm/node-postgres";
 import { Pool } from "pg";
 
 import * as schema from "./schema";
+import { PGlite } from "@electric-sql/pglite";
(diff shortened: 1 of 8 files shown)


packages/db/src/client.ts
```diff
@@ -12,7 +16,14 @@ export const createDrizzleClient = (): dbClient => {
   const connectionString = process.env.POSTGRES_URL;
 
   if (!connectionString) {
-    throw new Error("POSTGRES_URL environment variable is not set");
+    console.log("POSTGRES_URL environment variable is not set, using PGLite");
+
+    const client = new PGlite({ dataDir: "./pgdata", extensions: { uuid_ossp }});
+    const db = drizzlePgLite(client, { schema });
+
+    migrate(db, { migrationsFolder: "../../packages/db/migrations" });
```
hjball, line 24: We shouldn't need this @CyberL1 

It should be handled here: https://github.com/kanbn/kan/blob/main/apps/web/Dockerfile#L84

CyberL1: It'd be the ideal way as it'd be the most compatible, I know. But It seems this is the only way to migrate with the extension enabled. And it must be enabled at the time of the migrations, or it'd error otherwise:

hjball: Ah looks like a known issue: https://github.com/drizzle-team/drizzle-orm/issues/4476

We can remove once the fix is merged

hjball (approved):
LGTM

PR #139, opened 2025-08-10
fix: check if `POSTGRES_URL` env variable is set before migration

CyberL1:
Adds a check if the variable is set before running `pnpm db:migrate`. Just realized I forgot that in my previous PR


CyberL1:
Also added mkdir for pgdata in the image build cause it caused some weird permission errors when it didn't exist with volume

Issue #122, opened 2025-07-20
Possibility to change user password


maltsev-gorskij:
## ✨ Feature Request

**Is your feature request related to a problem? Please describe.**
After recent Kan fix invite links are back in action and working like a charm. But in fact after entering workspace, new user can't change password to perform futher login via login/password. For now new user invited via link can only relogin also by invite magic link. 

**Describe the solution you'd like**
- Users invited via link may be provided with screen to set their desired password, before they are redirected to the workspace, to be able in future to login with email and password pair
- In settings screen password change section must be provided for user to be able to change password as many times as they want

**Describe alternatives you've considered**
As alternative to provided solution, feature to change password may be implemented via context menu like this in `members' section
<img width="2176" height="220" alt="Image" src="https://github.com/user-attachments/assets/1513aeda-225b-4b9c-90bc-a8347d26d7d6" />
- Upon clicking on on this three dots, context menu with `Change password` for current user may be provided
- Upon clicking on `Change password` button, popup with password entering textfield fill appear

**Additional context**
-

**Implementation ideas (optional)**
-

**Would you like to work on this feature?**
- [ ] Yes, I'd like to help implement this feature
- [X] No, I'm just suggesting the feature

Eshwar1212-maker:
@maltsev-gorskij Hey can you asssign this to me? I would love to work on this

PR #145, opened 2025-08-15
feat: add change password feature

22vedant:
Implements the feature to update the user's password.

Tries to implement #122 

https://github.com/user-attachments/assets/c8ddc30a-5540-41eb-a7fa-e4860e84cc78


hjball:
This is a great start @22vedant 

We should extend this functionality to:

- Display a "Forgotten password?" message on the login page
- Clicking on it prompts user to enter their email (can prefill if user has provided email already)
- Password reset email is sent: https://www.better-auth.com/docs/authentication/email-password#request-password-reset
- User clicks on the URL in the email and updates their password
- Authenticated users is redirected to the `/boards` page

However this solves the immediate issue so let's merge and follow up separately

hjball (approved):

PR #149, opened 2025-08-17
fix: board not found flicker on reloading public boards

Rythm18:
Fixes:
- Board not found verbiage is shown for a second when reloading public board page

- Using next-theme's ThemeProvider to handle dark/light modes (handling the flickerof light mode on reloading)

**theme.tsx can be removed if we can move forwatrd next-theme's ThemeProvider.**


Rythm18:
@hjball fixed!

hjball:
Looks like the background is pulling through correctly, but the hero image is still using the light variant on first load @Rythm18  

Screenshot 2025-08-20 at 12 10 48

PR #185, opened 2025-09-19
feat(web): added image cropping to avatar uploads

nahSystemu:
The issue:
image

After realizing their is an issue with users uploading avatars that aren't square, with a 1:1 ratio I remembered that I seen an image crop component when I was researching a project for myself. 

> https://github.com/sujjeee/shadcn-image-cropper/blob/main/src/components/image-cropper.tsx
Used this as reference

After a while of struggling with understanding how the project is structured and setup for local development, I got this working on my local environment with minIO for S3. 

Here is a preview of it working as intended, but do check if everything suffices and there aren't any edge cases that would break something.

https://github.com/user-attachments/assets/f1d78aa6-5090-4fb3-87b4-4b02a6c6121f


hjball:
Great work @nahSystemu

hjball (approved):

PR #186, opened 2025-09-19
feat: added docs to website

nahSystemu:
After some research and not being able to directly find resources in the documentation itself, I did some digging and searching through issues and the discord. I was able to find a guide and some instructions on how to get minIO working with the project. 

After getting my instance working I decided to document it and actually add dedicated documentation for users who might want to do it also. 

Whilst doing this I moved also added the basic self-host guide from the readme so it's more centralised I guess. Hopefully some of this helps


hjball (approved):
Great stuff @nahSystemu

PR #190, opened 2025-09-20
fix: bulkCreate to stop imports with duplicate indices

nahSystemu:


hjball (approved):
Nice work @nahSystemu

PR #191, opened 2025-09-20
additions for #190 with similar checks for when working with cards

nahSystemu:


packages/db/src/repository/card.repo.ts
```diff
@@ -606,9 +713,53 @@ export const reorder = async (
...
+            FROM "card"
+            WHERE "listId" = ${affectedListIds[0]} AND "deletedAt" IS NULL
+          )
+          UPDATE "card" c
+          SET "index" = o.new_index
+          FROM ordered o
+          WHERE c.id = o.id;
+        `);
+      } else if (affectedListIds.length === 2) {
+        await tx.execute(sql`
```
hjball, line 734: We could probably just always default to this query and remove the one above, but not a blocker to this getting merged

hjball (approved):
Nice work @nahSystemu

PR #193, opened 2025-09-23
Traduções em pt-br

doug-fsg:


nahSystemu:
So your way of implementing Portuguese was by removing Russian? Can you fix the removal of a previously added language?

hjball:
Resolved in: https://github.com/kanbn/kan/pull/276

PR #194, opened 2025-09-24
feat: real-time events initial implementation

nahSystemu:
This implements SSE events to get a some what real-time updating feel. Where you instantly see cards updated on a board, information updated live inside the card and so on. 

Very rough implementation, needs testing, have been using a build with this implementation on my personal self-hosted instance. So far so good, no errors or issues, everything seems to work.

Issue: https://github.com/kanbn/kan/issues/7


hjball:
Sounds like we're in agreement on websockets instead of SSE events so I'll close this for now

Issue #240, opened 2025-11-12
Filter Board by Lists


SoniArpit:
## ✨ Feature Request

**Is your feature request related to a problem? Please describe.**
When working with boards that have many lists, users need a way to focus on specific lists by filtering the board view.

**Describe the solution you'd like**
Add a "Lists" filter option in the board filter dropdown (similar to existing Members and Labels filters) that allows users to:

- Select multiple lists from a multi-select dropdown
- Show only the selected lists on the board
- Clear the filter to show all lists again

**Additional context**

- Extends existing filter functionality (members and labels)
- Filter should persist in URL query parameters
- When no lists are selected, all lists are displayed (default behavior)

**Would you like to work on this feature?**

- [x] Yes, I'd like to help implement this feature
- [ ] No, I'm just suggesting the feature

hjball:
Good suggestion @SoniArpit, I've assigned you to the issue. Feel free to drop me any questions here or in the [discord](https://discord.com/invite/e6ejRb6CmT) 👌

hjball:
Resolved: https://github.com/kanbn/kan/pull/241

PR #241, opened 2025-11-13
feat: add list filter to board view

SoniArpit:
## Description
Adds list filtering to board view - users can now filter which lists are displayed, similar to existing member and label filters.

## Changes
- Added Lists filter to Filters component
- Multi-select filtering with URL persistence

issue: #240

### DIFF
--- a/apps/web/src/views/board/index.tsx
+++ b/apps/web/src/views/board/index.tsx
@@ -420,13 +420,7 @@ export default function BoardPage({ isTemplate }: { isTemplate?: boolean }) {
                     members={boardData.workspace.members.filter(
                       (member) => member.user !== null,
                     )}
-                    lists={
-                      (
-                        boardData as typeof boardData & {
-                          allLists: typeof boardData.lists;
-                        }
-                      ).allLists
-                    }
+                    lists={boardData.allLists}
                     position="left"
                     isLoading={!boardData}
                   />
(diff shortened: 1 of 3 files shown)


packages/db/src/repository/board.repo.ts
```diff
@@ -70,6 +70,7 @@ export const getByPublicId = async (
   filters: {
     members: string[];
     labels: string[];
+    lists: string[];
```
hjball: You'll need add lists to the filters in the query

hjball (changes requested):
Great start @SoniArpit, but filters should be API-driven rather than filtered in the front end

apps/web/src/views/board/index.tsx
```diff
@@ -489,78 +492,88 @@ export default function BoardPage({ isTemplate }: { isTemplate?: boolean }) {
...
-                            list={list}
-                            setSelectedPublicListId={(publicListId) =>
-                              setSelectedPublicListId(publicListId)
-                            }
-                          >
-                            <Droppable
-                              droppableId={`${list.publicId}`}
-                              type="CARD"
+                        {boardData.lists
+                          .filter((list) => {
```
hjball: we can remove this and let the API handle it

hjball (approved):
Nice work @SoniArpit

Issue #231, opened 2025-10-28
[Bug] Disabling email does not remove magic link option on login page


simonliu99:
## 🐛 Bug Report

**Describe the bug**
Disabling email with `NEXT_PUBLIC_DISABLE_EMAIL: true` does not remove magic link option on login screen.

**To Reproduce**

1. Set docker environment variable `NEXT_PUBLIC_DISABLE_EMAIL: true`
2. Start Kan
3. Navigate to `https://localhost:3000`
4. See persistent magic link option

**Expected behavior**
Setting `NEXT_PUBLIC_DISABLE_EMAIL: true` should remove the magic link option per #188 and #195 

**Environment (please complete the following information):**
- OS: Docker on MacOS
- Browser Chrome
- App version/commit: v0.4.7

nahSystemu:
Looking into it

nahSystemu:
`NEXT_PUBLIC_DISABLE_EMAIL=true`
I get this result:
<img width="613" height="585" alt="Image" src="https://github.com/user-attachments/assets/be115e00-6b2e-4d6a-b16b-ca0220f1cb87" />

<img width="555" height="523" alt="Image" src="https://github.com/user-attachments/assets/a773384f-d88a-4e7d-8aa0-7f18e2ff74e2" />

Whilst with: `NEXT_PUBLIC_DISABLE_EMAIL=false`

<img width="551" height="486" alt="Image" src="https://github.com/user-attachments/assets/59597080-40b7-42ec-8a2d-93966da7f73b" />

Could you please provide steps to reproduce? 
Your `docker-compose.yml`, or anything relevant that would shine some light on the issue

simonliu99:
My `compose.yaml` is pretty basic:
```
services:
  kan-web:
    image: ghcr.io/kanbn/kan:0.4.7
    container_name: kan-web
    ports:
      - "3000:3000"
    environment:
      NEXT_PUBLIC_BASE_URL: https://kan.example.com
      BETTER_AUTH_SECRET: <secret>
      POSTGRES_URL: postgresql://kan:kan@postgres:5432/kan_db
      NEXT_PUBLIC_ALLOW_CREDENTIALS: false
      NEXT_PUBLIC_DISABLE_SIGN_UP: false
      NEXT_PUBLIC_DISABLE_EMAIL: true
      OIDC_CLIENT_ID: <id>
      OIDC_CLIENT_SECRET: <secret>
[...]

PR #244, opened 2025-11-14
fix(web): logic of disabling credential login

nahSystemu:
If credentials are disabled, then don't show the email login form at all. If credentials are disabled and no social providers are present, show message saying none are available.

Resolving issue: #231


nahSystemu:
@hjball Needs all language translations, but logic works

hjball (approved):
Nice one @nahSystemu

PR #255, opened 2025-11-26
feat: add support for keyboard shortcuts

sjsimpson:
## Summary

Adds a keyboard shortcuts system with support for single key presses and multi-key sequences.

### Features

- **Shortcut registration** via `useKeyboardShortcut` hook
- **Sequence support** (e.g., `g` then `h` for "go home")
- **Conflict detection** in development mode
- **Keyboard shortcuts legend** modal (press `?` to open)
- **Grouped shortcuts** (General, Navigation, Actions)
- **Tooltip component** for displaying shortcuts on hover

### Usage

```tsx
// Register a shortcut and get tooltip content
const { keys, tooltipContent } = useKeyboardShortcut({
  type: "PRESS",
  stroke: { key: "k", modifiers: ["META"] },
  action: () => openCommandPalette(),
  description: "Open command palette",
  group: "GENERAL",
});

// Use with Tooltip component
<Tooltip content={tooltipContent}>
  <Button>Search</Button>
</Tooltip>

// Or use keys directly
<span>Press {keys} to search</span>
```

Sequences work similarly:

```ts
useKeyboardShortcut({
  type: "SEQUENCE",
  strokes: [{ key: "g" }, { key: "h" }],
  action: () => navigate("/home"),
  description: "Go to home",
  group: "NAVIGATION",
});
```
### New Files

- `apps/web/src/providers/keyboard-shortcuts.tsx` - Core provider, hook, and legend modal
- `apps/web/src/components/Tooltip.tsx` - Reusable tooltip using tippy.js

### Updated Files

- `apps/web/src/styles/globals.css` - Added tooltip theme styles
- `apps/web/src/components/WorkspaceMenu.tsx` - Updated command palette shortcut (⌘K)
- `apps/web/src/views/boards/index.tsx` - Added create board/template shortcut (C)
- `apps/web/src/views/board/index.tsx` - Added create list shortcut (C)

### DIFF
--- a/apps/web/src/components/UserMenu.tsx
+++ b/apps/web/src/components/UserMenu.tsx
@@ -10,6 +10,7 @@ import { twMerge } from "tailwind-merge";
 import { authClient } from "@kan/auth/client";
 
 import { useIsMobile } from "~/hooks/useMediaQuery";
+import { useKeyboardShortcuts } from "~/providers/keyboard-shortcuts";
 import { useModal } from "~/providers/modal";
 import { getAvatarUrl } from "~/utils/helpers";
 
@@ -31,6 +32,7 @@ export default function UserMenu({
   const router = useRouter();
   const { theme, setTheme } = useTheme();
   const { openModal } = useModal();
+  const { openLegend } = useKeyboardShortcuts();
   const isMobile = useIsMobile();
 
   const handleLogout = async () => {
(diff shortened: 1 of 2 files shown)


hjball:
This is great, nice work @sjsimpson 🙌

I've made a few small changes to UI/UX, but the foundations are looking really solid. Great job. 

Couple small things:
- We should launch with a create card hotkey (probably the most common use case)
- I can't see create list in the shortcuts legend? **Edit:** Just realised it's the shortcuts you have available to you - nice

sjsimpson:
@hjball awesome! I looked through your changes and they are really nice UI/UX/DX improvements.

> We should launch with a create card hotkey (probably the most common use case)

I had thought about this, but it looks like there isn't currently a flow for "Create a card and assign it to a table", which I think would be most suitable for a global hotkey. I think it is for sure something that we would want, but figured this could launch and support the existing feature set and be expanded to add hotkeys as actions are made available.

> Just realised it's the shortcuts you have available to you - nice

Yeah! They are tied to the component that renders them, so individual components can each render their own hotkeys. The problem with doing this for a "create card on list" type of hotkey is that we would need to tie it somehow to the list name, index, or id. I don't think that's a very fun way of doing it.

Another thing I saw in libraries like `react-hotkey-hook` (that is currently not supported by this feature) is to support "scoped" hotkeys, so when you are in a list scope (focus is in some part of the list) `C` creates a list. Could be nice for keyboard navigation, but I still think the global "create a card and assign it to a list" is just butter UI/UX

What do you think? **Edit:** Should I try and add a full "create card in list" feature here? or is that a follow-up? Or a bad idea?

hjball:
It's a good point @sjsimpson, I think it makes sense to get this shipped and we can figure out how solve the create card hotkey as a separate PR. I wonder if we have `c` create the card and then digits for the lists e.g. `c3` creates a card in third list (needs more thought).

PR #263, opened 2025-12-01
feat: limit oidc domains

firegrass:
- Restrict user creation to allowed domains
- Use Google auth hinting


hjball (approved):
nice work @firegrass

PR #274, opened 2025-12-06
feat: paginate and merge frequent activities

Rythm18:
- Paginated activities, now it shows 20 activties + load more
- Merge frequent activities ( for description and labels for now )
- Showing count of times description updated in a time frame defined.
- Showing multiple labels added/removed updated in a defined time frame.


packages/api/src/utils/activities.ts
```diff
@@ -0,0 +1,129 @@
...
+  let i = 0;
+
+  while (i < activities.length) {
+    const current = activities[i];
+    if (!current) {
+      i++;
+      continue;
+    }
+
+    const isCountMergeable = MERGEABLE_COUNT_TYPES.includes(current.type as any);
```
hjball: You should be able to reference the activity types from the schema instead of using any on this line and the one below

packages/api/src/utils/activities.ts
```diff
@@ -0,0 +1,129 @@
+// Activity type from database query result
+type Activity = {
```
hjball: We can extend the return type of the query with the additional fields:

```
NonNullable<Awaited<ReturnType<typeof cardRepo.getPaginatedActivities>>
```

apps/web/src/views/card/components/ActivityList.tsx
```diff
@@ -53,7 +57,37 @@ const getActivityText = ({
...
+  mergedLabels?: string[];
 }) => {
+  const TextHighlight = ({ children }: { children: React.ReactNode }) => (
+    <span className="font-medium text-light-1000 dark:text-dark-1000">
+      {children}
+    </span>
+  );
+
+  if (type === "card.updated.description" && mergeCount && mergeCount > 1) {
+    return t`updated the description ${mergeCount} times`;
```
hjball: I don't think we need to show how many times it has been updated. Instead we should just show "User updated the description" once to keep the activity log from becoming too noisy

hjball:
Left a some comments but this is really great work @Rythm18  🙌

Rythm18:
> Left a some comments but this is really great work @Rythm18  🙌

Ahh will fix these asap!

Rythm18:
@hjball updated!

hjball:
Nice one @Rythm18, I'll take a look

hjball (approved):
Great work @Rythm18

PR #318, opened 2026-01-17
feat: add board favourites with animations

eliott-herbert-byrnes:
Added ability to favourite/unfavourite boards with visual indicators and 
smooth layout animations when boards reorder.

Changes:
- Add `favorite` boolean column to board schema with default false
- Generate database migration for favorite column
- Update board repository to include favorite field and sort by favorite status
- Update board API router to accept favorite parameter in update endpoint
- Add star icon UI that appears on hover (desktop) or always visible (mobile)
- Implement toggle favorite functionality with optimistic UI updates
- Add Framer Motion layout animations for smooth board reordering
- Boards now sort with favorites first, then alphabetically by name

Technical details:
- Schema: Added boolean favorite column with NOT NULL and default false
- Repository: Updated getAllByWorkspaceId to include favorite field and 
  sort by desc(favorite) then asc(name)
- API: Extended board.update input schema to accept optional favorite boolean
- UI: Star icon uses Tailwind group-hover for desktop, always visible on mobile
- Animations: Framer Motion layout prop enables smooth position transitions


eliott-herbert-byrnes:
I'm thinking it might be more appropriate to have the favourite star always visible when a board is favourited, rather than just on hover. Let me know what you think @hjball.

packages/db/migrations/20260117220951_zippy_nekra.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "board" ADD COLUMN "favorite" boolean DEFAULT false NOT NULL;
```
hjball: Not a biggie but best to give migrations a name with

```
drizzle-kit generate --name="AddFavouriteColumnToBoard"
```

hjball:
This is looking great @eliott-herbert-byrnes - big fan of the reordering animation!

Few bits of feedback:
- I think you're right - it would be better to always show the star once favourited
- Don't forget the ordering on the public board page (don't think we should show the star)
- Should we allow only workspace admins to favourite the board for now?
- I wonder if long term we need a junction table for user-board favourites? Something that would allow per user favourites, rather than global per board. What do you think?

eliott-herbert-byrnes:
> This is looking great @eliott-herbert-byrnes - big fan of the reordering animation!
> 
> Few bits of feedback:
> - I think you're right - it would be better to always show the star once favourited
> - Don't forget the ordering on the public board page (don't think we should show the star)
> - Should we allow only workspace admins to favourite the board for now?
> - I wonder if long term we need a junction table for user-board favourites? Something that would allow per user favourites, rather than global per board. What do you think?

Thanks for the feedback, I will work on those changes later this week. I also need to add the function to the board drop-down menu.

Re: permission based favourites: I hadn't considered it from an organisation perspective. I'm thinking along the same lines you are, per user favourites would definitely be the optimal choice in the long term. I'm happy to refactor and permission to admins for now.

eliott-herbert-byrnes:
Resolved based on feedback, slightly improved the colouring / aesthetic of the icon. Also, added functionality to board dropdown menu with pop ups.

hjball:
Thanks for making those changes - this is looking great @eliott-herbert-byrnes 

> I wonder if long term we need a junction table for user-board favourites? Something that would allow per user favourites, rather than global per board. What do you think?

Just thinking about it and I reckon it'll be better to make this change now rather than later (otherwise we're going to create more work for ourselves down the line)

To do this properly we should:
- Create a `user_board_favourite` junction table, something like:
```
export const userBoardFavorites = pgTable(
  "user_board_favorite",
  {
    userId: uuid("userId")
      .notNull()
      .references(() => users.id, { onDelete: "cascade" }),
    boardId: bigint("boardId", { mode: "number" })
      .notNull()
      .references(() => boards.id, { onDelete: "cascade" }),
    createdAt: timestamp("createdAt").defaultNow().notNull(),
  },
  (table) => [
    primaryKey({ columns: [table.userId, table.boardId] }),
    index("user_board_favorite_user_idx").on(table.userId),
    index("user_board_favorite_board_idx").on(table.boardId),
  ],
);
```
- Update the queries to join the favourites table
- Update the API to filter/join by userId
- Favouriting a board will add new entry to the table and unfavouriting will remove it

Appreciate this is a big change but I think it'll be absolutely worth it. Happy to jump in and make the changes too - just let me know

eliott-herbert-byrnes:
@hjball I worked on implementing user board favourites. Let me know what you think. One thing I did notice whilst testing, there is currently no rate limiting applied to board APIs, not a huge threat, but users can currently fave/unfave and do all crud board procedures unlimited times. I don't have a huge amount of expertise on the db side, but I thought it would be worth mentioning.

hjball:
Great work  @eliott-herbert-byrnes

One small request about the member assertion and a few conflicts that need sorting, but otherwise this is looking good to ship.

hjball:
Thanks for flagging the rate limiting too - this is urgently needed so i'll prioritise this for the next release

hjball:
Great work with this one @eliott-herbert-byrnes, sorry for the delay in getting it merged!

PR #323, opened 2026-01-19
Show display name instead of email

wrefgtzweve:
Currently it always shows your current email in the bottom left, it'd be best if this instead shows the users display name so any screenshots, screenshare etc wont display personal information. Email can still be seen through the members tab.

image

Also made the email show on the settings menu (censored in the screenshot):
image


hjball:
Not every user has a display name so we should fallback to email, otherwise these changes lgtm @wrefgtzweve

hjball:
Nice one @wrefgtzweve

PR #325, opened 2026-01-19
Add `NEXT_API_BODY_SIZE_LIMIT` config

wrefgtzweve:
Adds a env var `NEXT_API_BODY_SIZE_LIMIT` that allows users to up the default upload limit, currently it's 1mb (a nextjs default)


hjball:
I don't have any objections to this, but I'm curious to know what limits were you hitting and for which routes? @wrefgtzweve

hjball (approved):

wrefgtzweve:
It was in a card comment, i was trying to upload a small video of a example/idea
Might not hurt increasing the 1mb to something like ~10mb, i don't know if there's big downsides for nextjs though.

hjball:
Makes sense, I completely forgot about video uploads in cards (it's a bug really - we should be uploading the videos to s3 rather than storing it as a blob in pg)

hjball:
You'll also need to update:
- .env.example
- docker-compose.yml
- cloud/docker-compose.yml (not needed but would be helpful to keep it in sync)
- turbo.json

wrefgtzweve:
> Makes sense, I completely forgot about video uploads in cards (it's a bug really - we should be uploading the videos to s3 rather than storing it as a blob in pg)

Could make it optional, if S3 is present, put it there if not put it in pg
Would be nice for profile pictures too

hjball:
Thanks for making those changes @wrefgtzweve

PR #357, opened 2026-02-03
fix: board and workspace URL links and copy behavior

charitea:
- Use /boards/{id} for private board links so copied link works; keep pretty URL for public boards
- Show matching path in board header (pretty path for public, boards/id for private)
- Copy link button copies URL to clipboard instead of opening in new tab (opening a new tab seems redundant since we're already on it)
- Edit workspace URL button links to /settings/workspace instead of account settings

https://github.com/user-attachments/assets/d8ce6cb7-af9a-4263-8858-50fbae51b251


hjball (approved):
Nice work @charitea

Issue #341, opened 2026-01-28
Card title text not wrapped correctly


oonqt:
When the title text within a card exceeds the length of the card, the text continues off the card rather than wrapping. This only seems to happen to text without spaces... I found this when I set a card title to a long url

 
![Image](https://github.com/user-attachments/assets/d645d81a-63f5-4cec-9f27-f268b5f93167)

hjball:
Good spot @oonqt - will take a look into this one

PR #367, opened 2026-02-07
fix: prevent long card titles from overflowing column width

kOaDT:
## Summary

Long unbreakable strings (e.g. URLs) in card titles caused columns to expand beyond their fixed width, triggering horizontal scrolling and degrading the drag-and-drop experience

Added overflow-hidden to the card container and break-words to the title span so long strings wrap naturally within the card boundaries

Linked issue: https://github.com/kanbn/kan/issues/341

### Before

Capture d’écran du 2026-02-07 22-58-56

### After

Capture d’écran du 2026-02-07 22-58-41

## Test plan

- [ ] Create a card with a long URL as the title
- [ ] Verify the title wraps within the card without expanding the column width
- [ ] Verify no horizontal scrollbar appears on the board
- [ ] Verify drag-and-drop still works correctly with long titles
- [ ] Verify normal short titles are unaffected

---

Thank you for building and maintaining Kan. It's a beautifully crafted OSS. Happy to contribute! 🙏

### DIFF
--- a/apps/web/src/views/board/components/Card.tsx
+++ b/apps/web/src/views/board/components/Card.tsx
@@ -66,8 +66,8 @@ const Card = ({
   const hasDueDate = !!dueDate;
 
   return (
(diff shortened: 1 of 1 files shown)


hjball:
Glad to have you onboard!

PR #369, opened 2026-02-08
feat: add two types of roadmap templates

program-the-brain-not-the-heartbeat:
Add two different types of roadmap templates.

## 1. Basic Roadmap

Similar to the [official Kan Roadmap](https://kan.bn/kan/roadmap) but slightly modified.

- `Requested`
- `Planned`
- `In Progress`
- `Done`

## 2. Extended Roadmap
- `Requested`
- `Under Review`
- `Planned`
- `In Progress`
- `Done`
- `Rejected`

### DIFF
--- a/apps/web/src/views/boards/components/TemplateBoards.tsx
+++ b/apps/web/src/views/boards/components/TemplateBoards.tsx
@@ -24,6 +24,18 @@ export const getTemplates = (): Template[] => [
     labels: [t`Bug`, t`Feature`, t`Enhancement`, t`Critical`, t`Documentation`],
   },
   {
+    id: "roadmap-basic",
+    name: t`Basic Roadmap`,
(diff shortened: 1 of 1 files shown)


program-the-brain-not-the-heartbeat:
firefox_PX21dNAv8SPhmSrX

hjball:
I wonder if we just make this the default roadmap template. What do you think? @program-the-brain-not-the-heartbeat

program-the-brain-not-the-heartbeat:
I currently use both the basic roadmap (personal projects) and the extended (work) but I don't want to get too carried away by adding lots of templates into the codebase.

The extended version probably makes the most sense if we have to limit it to 1, as this will allow most users who just want a quick template without having to manually create the cards and keeps activity in "Rejected" too for ease of viewing/history of why cards were declined.

PR #370, opened 2026-02-08
fix: handle x-forwarded-host and proxy issues behind Cloudflare

program-the-brain-not-the-heartbeat:
This solves issues with Cloudflare setup and comma separated for both `X-Forwarded-Host` and `X-Forwarded-Proto` HTTP headers for proxies.


hjball (approved):
Looks sensible to me @program-the-brain-not-the-heartbeat

PR #371, opened 2026-02-08
Join Workspace Email improvements

program-the-brain-not-the-heartbeat:
This is the last one for me this weekend I swear.

- This fixes a join workspace email bug due to checking if the url contains type=invite, but failed on percent-encoded URLs. This will now work on both.
- Renames some strings from "kan.bn" to "Kan" for consistency
- Respects the `NEXT_PUBLIC_WHITE_LABEL_HIDE_POWERED_BY` environment variable for this email message only.

The other templates should be reviewed for the last two list items but I will open a separate PR for that, possibly next weekend.

### DIFF
--- a/packages/auth/src/plugins.ts
+++ b/packages/auth/src/plugins.ts
@@ -225,10 +225,16 @@ export function createPlugins(db: dbClient) {
             },
           );
         } else {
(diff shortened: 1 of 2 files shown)


hjball (approved):
Great effort this weekend @program-the-brain-not-the-heartbeat 

Loving all these PRs 🙌

PR #383, opened 2026-02-11
feat: add close button to card view header

charitea:
Super simple, I found returning to the board a little annoying w/o a dedicated close button in card view. 
image


hjball:
You can return via the breadcrumb nav "Test board", but I don't think that's obvious so this seems like a good addition @charitea

PR #384, opened 2026-02-11
feat: add easy assigned filter

charitea:
image

Resolves 1. of https://discord.com/channels/1346412950641967134/1450160176995766292


hjball:
I think for that specific issue, we want some way of filtering for cards assigned to me across multiple boards @charitea 

We already have member filters for filtering on a single board (although we could update the name of the currently authenticated user to be "Assigned to me" to make it clearer)

charitea:
Ohhhh that makes sense. I'll try and come up with a solution for that then

hjball:
Closing this one for now @charitea

PR #386, opened 2026-02-11
feat:archived boards

eliott-herbert-byrnes:
- Boards are now split in to two lists 'Active | Archived'. 
-  Archived/Unarchived is activated through the boardDropdown.
- Desktop follows same UX as settings tabs, mobile is optimised to a dropdown menu.
- added isArchived column, & board_is_archived_idx index on the board table.

### DIFF
--- a/packages/db/migrations/meta/_journal.json
+++ b/packages/db/migrations/meta/_journal.json
@@ -187,15 +187,15 @@
     {
       "idx": 26,
       "version": "7",
-      "when": 1772049587901,
-      "tag": "20260225195947_AddIsArchivedToBoard",
+      "when": 1771192000000,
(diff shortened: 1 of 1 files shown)


eliott-herbert-byrnes:
> Thanks for your patience with this one @eliott-herbert-byrnes
> 
> It's looking great, just a couple of requested changes:
> 
> * Revert the changes under `/locales` (handled on merge now)
> * Replace the archive and unarchived endpoints with `update`
> 
> You'll also need to reorder the migrations (easiest way to do this is just to remove your entry from `_journal`, delete your snapshot and migration file, then regenerate). This is becoming a common blocker so I'm going to look into a workflow to handle migration conflicts automatically)

Should be all sorted! lmk if there are any further changes needed.

PR #391, opened 2026-02-14
feat(db): add webhook schema, migration, and repository

nickmeinhold:
## Summary
- Add `workspace_webhooks` table with migration (event enum, URL, secret, event subscriptions, active flag)
- Add webhook repository with CRUD operations (create, update, get, list, delete)
- Extend card and list repos to return board/list names for webhook payload context

This is the database foundation for the webhooks feature, split from #343 for easier review.

## PR series
- **PR 1 (this)**: DB schema & repository
- PR 2: Webhook delivery utility + card event integration
- PR 3: Webhook CRUD API router + tests
- PR 4: Webhook management UI

## Test plan
- [ ] Migration runs cleanly on a fresh database
- [ ] Webhook CRUD operations work via repository methods
- [ ] Card/list repo changes return board and list names correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #392, opened 2026-02-14
feat(api): add webhook delivery utility and card event integration

nickmeinhold:
## Summary
- Add `sendWebhookToUrl()` with HMAC-SHA256 signature, `X-Webhook-Event`/`X-Webhook-Timestamp`/`X-Webhook-Signature` headers, and 10s abort timeout
- Add `sendWebhooksForWorkspace()` for fan-out delivery to all matching active webhooks (fire-and-forget via `Promise.allSettled`)
- Add `createCardWebhookPayload()` helper for building structured payloads with card, board, list, and user context
- Fire webhooks in card `create`, `update`, `move`, and `delete` mutations
- Add unit tests for all utility functions

## Dependencies
Depends on #391 (DB schema & repository)

## PR series
- PR 1: DB schema & repository (#391)
- **PR 2 (this)**: Webhook delivery utility + card event integration
- PR 3: Webhook CRUD API router + tests
- PR 4: Webhook management UI

## Test plan
- [ ] Unit tests pass for webhook utility functions
- [ ] Card create/update/move/delete fires webhooks to active subscriptions
- [ ] Failed webhook deliveries are logged but don't block card operations
- [ ] HMAC signatures are correct and verifiable by consumers

### DIFF
--- a/packages/api/src/utils/webhook.test.ts
+++ b/packages/api/src/utils/webhook.test.ts
@@ -9,6 +9,7 @@ import {
   sendWebhookToUrl,
   sendWebhooksForWorkspace,
   createCardWebhookPayload,
+  webhookUrlSchema,
   type WebhookPayload,
 } from "./webhook";
 
@@ -403,6 +404,12 @@ describe("webhook utilities", () => {
 
       await sendWebhooksForWorkspace(mockDb, 1, mockPayload);
 
+      // Event filtering now happens at DB level
+      expect(mockGetActiveByWorkspaceId).toHaveBeenCalledWith(
+        mockDb,
+        1,
+        "card.created",
+      );
       expect(global.fetch).toHaveBeenCalledTimes(2);
       expect(global.fetch).toHaveBeenCalledWith(
         "https://example.com/webhook1",
@@ -414,20 +421,17 @@ describe("webhook utilities", () => {
       );
     });
 
-    it("does not send to webhooks not subscribed to the event", async () => {
-      mockGetActiveByWorkspaceId.mockResolvedValueOnce([
-        {
-          id: 1,
-          publicId: "wh-1",
-          url: "https://example.com/webhook1",
-          secret: null,
-          events: ["card.deleted"], // Not subscribed to card.created
-          active: true,
-        },
-      ]);
+    it("does not send when no webhooks match the event (DB-level filtering)", async () => {
+      // DB-level event filter returns empty array when no webhooks match
+      mockGetActiveByWorkspaceId.mockResolvedValueOnce([]);
 
       await sendWebhooksForWorkspace(mockDb, 1, mockPayload);
 
+      expect(mockGetActiveByWorkspaceId).toHaveBeenCalledWith(
+        mockDb,
+        1,
+        "card.created",
+      );
       expect(global.fetch).not.toHaveBeenCalled();
     });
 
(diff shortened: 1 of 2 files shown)


MaxwellMergeSlam (commented):
## MaxwellMergeSlam's Review

**Verdict:** REQUEST_CHANGES

**Summary:** The webhook delivery utility is clean and well-tested, but the card router has a bug where `boardId` is actually the workspace ID, and there's an SSRF vector in the URL handling.

The Terminator: *"I'll be back."* — And so will this PR, once you fix the boardId bug that's been staring you in the face since line 1.

**Findings:**

- **`packages/api/src/routers/card.ts:35` and `:89` and `:136`**: `boardId: String(list.workspaceId)` / `boardId: String(card.workspaceId)` — **THIS IS A BUG.** The `boardId` field in the webhook payload is set to the WORKSPACE ID, not the actual board ID. Every webhook consumer receiving this payload will get wrong data in `data.card.boardId` and `data.board.id`. The actual board ID is available on the list/card objects but isn't being used.

- **`packages/api/src/utils/webhook.ts:56`**: The `sendWebhookToUrl` function accepts ANY URL and makes a POST request to it. This is a **Server-Side Request Forgery (SSRF) vector**. An admin can configure a webhook pointing to `http://169.254.169.254/latest/meta-data/` (AWS metadata), `http://localhost:5432` (internal DB), or any internal service. Mitigations to consider: validate that URLs are HTTPS-only, block private/internal IP ranges, or use an allowlist. At minimum, add a comment acknowledging this is admin-only and the risk is accepted.

- **`packages/api/src/utils/webhook.ts:53`**: The `AbortController` timeout with `setTimeout` — the timeout ID is properly cleared in both success and error paths, which is good. But the 10-second timeout is hardcoded with no way to configure it. Fine for now, but worth a named constant.

- **`packages/api/src/routers/card.ts:53`**: The `webhookChanges` diff object checks `input.title` truthily (`if (input.title && ...)`), which means setting a title to an empty string would skip the change detection. This might be intentionally prevented by validation elsewhere, but it's worth noting.

**The Good:**

- Fire-and-forget pattern with `void sendWebhooksForWorkspace(...)` is exactly right — webhook delivery should never block card operations
- `Promise.allSettled` for parallel delivery means one slow/failing webhook doesn't block others
- HMAC-SHA256 signature implementation is standard and correct
- Test coverage is excellent — 429 lines of tests for 175 lines of implementation, covering happy paths, failures, timeouts, and event filtering
- The `createCardWebhookPayload` helper keeps the card router clean

**The Concerns:**

- The boardId bug is a data integrity issue — every payload sent will have incorrect board identification
- SSRF risk is real for self-hosted deployments behind firewalls
- No retry mechanism — fire-and-forget is fine for v1, but there's no visibility into delivery failures beyond a `console.error`

Neo: *"I know kung fu."* — Yeah, well this code doesn't know its boardId from its workspaceId. Fix that and we're golden.

KelvinBitBrawler (commented):
## KelvinBitBrawler's Review

**Verdict:** REQUEST_CHANGES

**Summary:** Implementing asynchronous fire-and-forget logic without a safety net is like stepping out of the airlock without checking the seal—thermodynamically unsound and likely to end in a vacuum-induced mess.

**Findings:**

- **packages/api/src/routers/card.ts:172, 1059, 1154**
  You're using `void sendWebhooksForWorkspace(...)` to fire these off into the void. That's cute, but `sendWebhooksForWorkspace` is `async` and awaits a database call (`getActiveByWorkspaceId`) before it hits the `Promise.allSettled` block. If that DB call throws (connection timeout, pool exhaustion, entropy death of the universe), you're creating an **Unhandled Promise Rejection**.
  MacReady: *"If we've got any surprises for each other, I don't think we're in much shape to handle them."*
  **Fix it:** Wrap the body of `sendWebhooksForWorkspace` in a `try/catch` block so it absorbs failures like a black hole, or chain a `.catch()` in the router. Don't let your laziness crash the process.

- **packages/api/src/utils/webhook.ts:102**
  `sendWebhooksForWorkspace` implementation.
  The filtering of events happens in memory (Line 112). For now, the heat generated by this inefficiency is negligible, but if a workspace has 10,000 webhooks, you're going to spike the CPU temperature. I'll let it slide for now, but watch your delta-T.

**The Good:**
- **packages/api/src/utils/webhook.ts:64-73**
  Using `AbortController` for the 10s timeout. Most devs forget that requests can hang until the heat death of the universe. You didn't. Good.
- **packages/api/src/utils/webhook.ts:40**
  HMAC signature generation on the stringified body. Cryptographically sound. You passed the Turing test on this one.

**The Concerns:**
- **packages/api/src/routers/card.ts:1136**
  You're fetching the full card (`getByPublicId`) right before deleting it just to get data for the webhook. It's an extra DB roundtrip that increases the system's total entropy, but I suppose necessary since you didn't pass the card data in from the client. Acceptable, but inefficient.

**Final thoughts:**
Your error handling boundaries are non-existent. Fix the unhandled rejection risk, or I'll freeze this PR at absolute zero.

nickmeinhold:
Pushed review feedback changes to this branch (d433b2f):
- Extracted URL validation into reusable `webhookUrlSchema` zod validator for SSRF checks
- Wrapped `sendWebhooksForWorkspace` in try/catch to prevent unhandled promise rejections
- Documented SSRF risk mitigation on `sendWebhookToUrl`
- Added corresponding tests

PR #409, opened 2026-02-21
feat: added board ordering in project using dnd

kotserge:
This pull request contains feature for reordering boards by drag-and-drop through the UI:

  - Added a position column to the board schema to enable explicit ordering within a workspace
  - Implemented position management in the repository layer: auto-assign on create, decrement on delete, CASE-based reorder with duplicate auto-heal
  - Added position field to the board update tRPC mutation to trigger reordering
  - Built drag-and-drop UI using @dnd-kit with rectSortingStrategy for CSS grid support, programmatic navigation to prevent click-on-drop, and optimistic cache updates with rollback
  on error


packages/db/src/schema/boards.ts
```diff
@@ -57,6 +58,7 @@ export const boards = pgTable(
     visibility: boardVisibilityEnum("visibility").notNull().default("private"),
     type: boardTypeEnum("type").notNull().default("regular"),
     sourceBoardId: bigint("sourceBoardId", { mode: "number" }),
+    position: integer("position").notNull(),
```
hjball: This will update the position for all workspace members. Ideally we want to allow each user to customise the order of their boards so we'll need to do something similar to `user_board_favourites`.

In fact we should probably merge `user_board_favourites` into `user_board` and store `position` and `isFavourite` on there. That'll give us greater flexibility in future.

We'll need to:
- Create `board_user` table with `userId`, `boardId`, `position`, `isFavourite`
- Create migration to create a `board_user` record for each board a user belongs (all in every workspace they belong to) and check against the `board_user_favourites` table to set `isFavourite`
- Adding a new board should create a `board_user` record
- Updating the `position` should reorder the `board_user` records
- Deleting a board should remove any `board_user` records for that board
- Favouriting/unfavouriting a board should update the `board_user` record
- Boards should be ordered by their position/isFavourite for a given user

Sorry for the dump - there's quite a lot to do here so I'm very happy to help. Just give me a shout if you've got any questions

hjball (changes requested):
Good start @kotserge - I've left some feedback. Give me a shout if you've got any questions

kotserge:
Hey @hjball. 

My bad, I let this drag on for so long. While I am not using this anymore (as I do not need _multi-tenant features_ and a local file is enough for me) I do not want to let our time go to waste. 

I have incorporated your suggestions and this should now work on a per-user basis. The changes introduce: 
- replaced the `userBoardFavorites` table with `boardUsers` (`board_user`) (including migration)
  - `userId` (uuid, FK -> users, cascade), `boardId` (bigint, FK -> boards, cascade)
  - `position` (integer, not null), `isFavourite` (boolean, not null, default false)
  - `createdAt` (timestamp, default now)
  - Composite PK `(userId, boardId)`; indexes `board_user_user_idx`, `board_user_board_idx`
- basic extensions of `repository`, `api`, `hooks` to handle member-joining and reorders 
- frontend adjustments to allow board card drag-n-drop
  - dnd-kit as new dependency for functionality

PR #418, opened 2026-02-25
fix: allow invited users to sign up when registration is disabled

### LINKED ISSUE #411: Allow disabling registration completely while still allowing login
## ✨ Feature Request

**Is your feature request related to a problem? Please describe.**
There is no way to disable registration features. Which can lead to unauthorized registration on public instances since there is no administration dashboard for controlling that.

**Describe the solution you'd like**
Allow controlling through a flag if an instance accepts registration or not.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

**Implementation ideas (optional)**
Use an environment variable like `REGISTRATION_ENABLED=true` in order to control who can access this instance when not using OIDC.

**Would you like to work on this feature?**

- [ ] Yes, I'd like to help implement this feature
- [x] No, I'm just suggesting the feature

nickmeinhold:
## Summary

Fixes #411 — when `NEXT_PUBLIC_DISABLE_SIGN_UP=true`, invited members are blocked from signing up because:

1. **better-auth's `disableSignUp` flag** rejects email/password sign-ups at the framework level, *before* the database hook can check for pending invitations
2. **The signup page** renders a hard "Sign up disabled" dead end, even when the user arrived via an invite link

### Changes

- **`packages/auth/src/auth.ts`** — Set `disableSignUp: false` in the better-auth config. The sign-up restriction is already handled by the `user.create.before` database hook in `hooks.ts`, which checks `memberRepo.getByEmailAndStatus()` for a pending invitation and only allows the sign-up through if one exists.
- **`apps/web/src/views/auth/signup/index.tsx`** — Detect invite flows (`?next=/invite/...`) and show the sign-up form even when registration is globally disabled.
- **`apps/web/src/views/auth/login/index.tsx`** — Show the "Sign up" link when the user is in an invite flow, so they can navigate from login to signup.

### How it works

When `NEXT_PUBLIC_DISABLE_SIGN_UP=true`:
- **Public visitors** → signup page shows "Sign up disabled" (unchanged)
- **Invited users** arriving via `/invite/:code` → redirected to `/signup?next=/invite/:code` → signup form is shown → the database hook verifies the pending invitation and allows user creation

## Test plan

- [ ] Set `NEXT_PUBLIC_DISABLE_SIGN_UP=true`
- [ ] Verify visiting `/signup` directly shows "Sign up disabled"
- [ ] Invite a user to a workspace
- [ ] Open the invite link → click "Sign Up" → verify the signup form is shown
- [ ] Complete signup → verify the user is created and the invite is accepted
- [ ] Verify OIDC/social sign-ups still respect the flag (blocked without invite, allowed with invite)
- [ ] Verify normal operation when `NEXT_PUBLIC_DISABLE_SIGN_UP` is unset or `false`

### DIFF
--- a/packages/auth/src/hooks.test.ts
+++ b/packages/auth/src/hooks.test.ts
@@ -127,6 +127,54 @@ describe("createDatabaseHooks", () => {
       delete process.env.BETTER_AUTH_ALLOWED_DOMAINS;
     });
 
+    // The user.create.before hook fires for ALL sign-up paths including
+    // OIDC/social — verify invite bypass works regardless of auth method.
+    it("allows OIDC/social sign-up when disabled but user has a pending invitation", async () => {
+      mockEnv.mockImplementation((key: string) =>
+        key === "NEXT_PUBLIC_DISABLE_SIGN_UP" ? "true" : undefined,
+      );
+      const oidcUser = {
+        ...fakeUser,
+        id: "user-oidc",
+        email: "sso@corp.com",
+        image: "https://provider.com/avatar.jpg",
+      };
+      mockGetByEmailAndStatus.mockResolvedValue({
+        id: "member-2",
+        email: "sso@corp.com",
+        status: "invited",
+      });
+
+      const result = await hooks.user.create.before(oidcUser, {});
+      expect(result).toBe(true);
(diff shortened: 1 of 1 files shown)


hjball (approved):
Nice one @nickmeinhold, I didn't realise we were so close already

Issue #429, opened 2026-03-04
Font Variable for users


clemenko:
## ✨ Feature Request

**Is your feature request related to a problem? Please describe.**
The default font size are too small. I end up zooming in on the browser to 120%. It would be great to have a setting for small medium large in the user settings tab.

**Describe the solution you'd like**
The font size setting would change the default font size.

**Describe alternatives you've considered**
I zoom in within the browser. 

**Additional context**
default : 
![Image](https://github.com/user-attachments/assets/731a9b9d-a6d8-422c-81c7-a470f8c60704)

zoomed:
![Image](https://github.com/user-attachments/assets/521e8267-93eb-46e6-ad84-5204c9295c64)

**Would you like to work on this feature?**

- [ x ] Yes, I'd like to help implement this feature
- [ ] No, I'm just suggesting the feature

clemenko:
PR https://github.com/kanbn/kan/pull/430 create for this

clemenko:
pushed to clemenko/kan for testing.

hjball:
Closed by: https://github.com/kanbn/kan/pull/430

PR #430, opened 2026-03-04
Fontsize PR for Issue #429

clemenko:
This is adding a feature to allow the user to change the font size.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@ services:
     image: ghcr.io/kanbn/kan:latest
     container_name: ${CONTAINER_NAME:-kan-web}
     ports:
-      - "${WEB_PORT:-3000}:3001"
+      - "${WEB_PORT:-3000}:3000"
     networks:
       - kan-network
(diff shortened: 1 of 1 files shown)


clemenko:
pushed to clemenko/kan for testing.

clemenko:
@hjball Any possibility to add this?

hjball (changes requested):
This is great @clemenko 

Couple of bits of feedback:

- Medium should be the default (mine defaulted to small when testing)
- Can we move this to a select component in account settings? You should be able to reuse/copy the select from language

clemenko:
Sure. Let me work on this tonight.  TO confirm you want it moved to the settings/account page. to play devils advocate I put it next to the theme section because both arr visual elements.

hjball:
Yeah please - I'm keen to keep the user menu from becoming bloated and I don't think changing between font sizes will be common (you'd usually only do it once)

clemenko:
ok. good point on the frequency. Just commited the change.

PR #463, opened 2026-04-06
fix: use stable public ids for card webhook

### LINKED ISSUE #457: Card.moved webhook — data.list.id returns different values for the same list across events
## 🐛 Bug Report

**Describe the bug**
In `card.moved` webhook payloads, `data.list.id` returns different numeric IDs for the
same list across different events. The same list (e.g. "Backlog") can have `id=3` in one
event and `id=5` in another.

**To Reproduce**
1. Configure a webhook for `card.moved` events on a board with lists: Backlog, Ready, In Progress, Review, Done
2. Move a card to **Done** — note `data.list.id`
3. In a different session, move a card to **Backlog** — note `data.list.id`

Event 1 — card moved to Done:
```json
{
  "data": {
    "list": { "id": "5", "name": "Done" },
    "changes": { "listId": { "from": 6, "to": 5 } }
  }
}
```

Event 2 — card moved to Backlog (different session):
```json
{
  "data": {
    "list": { "id": "5", "name": "Backlog" },
    "changes": { "listId": { "from": 2, "to": 5 } }
  }
}
```

`data.list.id = 5` maps to `"Done"` in one event and `"Backlog"` in another on the same board.

**Expected behavior**
`data.list.id` should be a stable, unique identifier for each list that does not change
between sessions or restarts.

Suggested fix — use the stable `publicId` from the REST API instead of an internal numeric ID:
```json
"list": {
  "id": "fuoepumx7aog",
  "name": "Ready"
}
```

**Screenshots**
N/A

**Environment:**
- OS: Linux (self-hosted)
- Browser: N/A (webhook consumer)
- Node version: N/A
- App version/commit: 0.5.6

**Additional context**
The stable `publicId` for each list is already available via `GET /api/v1/boards/:boardId`
but is not used in webhook payloads. Using it would resolve both this issue and Issue #1
(inconsistency between `data.list.id` and `changes.listId`).

vvaswani:
Fixes #457

### DIFF
--- a/packages/api/src/routers/card.ts
+++ b/packages/api/src/routers/card.ts
@@ -181,7 +181,7 @@ export const cardRouter = createTRPCRouter({
             title: input.title,
             description: input.description,
             dueDate: input.dueDate ?? null,
-            listId: String(newCard.listId),
+            listId: list.publicId,
           },
           {
             boardId: list.boardPublicId,
@@ -906,9 +906,18 @@ export const cardRouter = createTRPCRouter({
       );
 
       let newListId: number | undefined;
+      let newList:
+        | {
+            id: number;
+            publicId: string;
+            name: string;
+            boardId: number;
+            index: number;
+          }
+        | undefined;
 
       if (input.listPublicId) {
-        const newList = await listRepo.getByPublicId(
+        newList = await listRepo.getByPublicId(
           ctx.db,
           input.listPublicId,
         );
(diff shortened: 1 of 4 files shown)


packages/api/src/routers/card.ts
```diff
@@ -1055,29 +1064,38 @@ export const cardRouter = createTRPCRouter({
...
-            ? "card.moved"
-            : "card.updated",
+          movedToNewList ? "card.moved" : "card.updated",
           {
             id: String(result.id),
             title: result.title,
             description: result.description,
             dueDate: result.dueDate,
-            listId: String(newListId ?? existingCard.listId),
+            listId: currentWebhookListPublicId,
```
hjball, line 1093: Ideally we should rename this to `listPublicId` but fine for now

hjball (approved):
Nice work @vvaswani

PR #475, opened 2026-04-20
fix: remove duplicate OIDC provider registration in getSocialProviders

louiechilli:
The getSocialProviders endpoint was manually pushing "oidc" to the  providers array when OIDC env vars are present, but Better Auth already  includes the OIDC provider in ctx.context.socialProviders when it is  configured. This resulted in ["oidc","oidc"] being returned from  /api/auth/social-providers, causing a duplicate provider registration  that led to a TypeError: Cannot read properties of null (reading 'id')  during the OIDC login callback, preventing all OIDC logins.

Fix: remove the manual providers.push("oidc") block as it is redundant.


hjball (approved):
Good spot @louiechilli

PR #484, opened 2026-04-30
fix: unable to change password when using magic link

### LINKED ISSUE #483: Unable to set/change password — password field in profile settings has no functionality
When a user registers via invitation or logs in via magic link, they are unable to set or change their password. The "Change Password" field in the profile/account settings appears to exist in the UI but has no functionality — submitting it does not update the password.

**Steps to Reproduce:**
1. Invite a user to a self-hosted kan.bn instance
2. User accepts invite and logs in via magic link
3. User navigates to profile/account settings
4. User attempts to set or change their password
5. Nothing happens — no error, no confirmation, password is not updated

**Expected Behavior:**
Users should be able to set or change their password via the profile settings, especially relevant for self-hosted instances where email/password login (NEXT_PUBLIC_ALLOW_CREDENTIALS=true) is the primary authentication method.

**Actual Behavior:**
The password change field appears to be non-functional. No feedback is given to the user.

**Environment:**
- Self-hosted via Docker Compose
- NEXT_PUBLIC_ALLOW_CREDENTIALS=true
- NEXT_PUBLIC_DISABLE_SIGN_UP=true (invite-only)
- SMTP configured and functional (magic links are delivered)

**Additional Context:**
Confirmed via Discord with @Morfixx @nahSystemu who indicated a fix is in progress. Reporting as requested.

nahSystemu:
# Fix: Magic link users can now set a password from their account

Closes #483

## What changed

Users who signed up or were invited via magic link were shown a "Change Password" option in account settings that had no working functionality — the form required a current password that didn't exist for these users.

This PR fixes that broken flow and adds a proactive prompt so magic link users are nudged to set a password after login.

## Changes

- **Database** — `GET /users/me` now returns two new boolean flags: `hasPassword` and `hasMagicLinkAccount`, derived from the `account` table
- **Account settings** — the password section now shows "Set Password" or "Change Password" based on `hasPassword`, and the form hides the current password field when there is nothing to verify against
- **Post-login prompt** — magic link users with no password see a modal on their first page load after login, offering to set a password so they can use email/password login in future without requesting a magic link each time
- **Safety guard** — if the prompt surfaces for a user who already has a password (e.g. stale session state), the form is replaced with an informational message and a close button only

## Testing

- Log in via magic link with an account that has no password → prompt should appear
- Dismiss the prompt → should not reappear within the same browser session
- Set a password via the prompt → success toast, prompt never appears again on subsequent logins
- Navigate to account settings as a magic link user → section should read "Set Password"
- Navigate to account settings as a credential user → section should read "Change Password" with the current password field present

### DIFF
--- a/apps/web/src/components/Dashboard.tsx
+++ b/apps/web/src/components/Dashboard.tsx
@@ -8,6 +8,7 @@ import {
   TbLayoutSidebarRightCollapse,
   TbLayoutSidebarRightExpand,
 } from "react-icons/tb";
+import { t } from "@lingui/core/macro";
 
 import { authClient } from "@kan/auth/client";
 
@@ -15,7 +16,10 @@ import { useClickOutside } from "~/hooks/useClickOutside";
 import { useModal } from "~/providers/modal";
 import { useWorkspace, WorkspaceProvider } from "~/providers/workspace";
 import { api } from "~/utils/api";
+import Button from "./Button";
+import Modal from "./modal";
 import SideNavigation from "./SideNavigation";
+import { ChangePasswordFormConfirmation } from "~/views/settings/components/ChangePasswordConfirmation";
 
 interface DashboardProps {
   children: React.ReactNode;
@@ -43,7 +47,7 @@ export default function Dashboard({
   hasRightPanel = false,
 }: DashboardProps) {
   const { resolvedTheme } = useTheme();
-  const { openModal } = useModal();
+  const { openModal, closeModal, modalContentType } = useModal();
   const { availableWorkspaces, hasLoaded } = useWorkspace();
   const router = useRouter();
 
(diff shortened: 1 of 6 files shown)


hjball (commented):
I'm not sure this approach will work:

https://better-auth.com/docs/authentication/email-password#update-password

Looks like previous password is a requirement of `/change-password`: https://better-auth.com/docs/authentication/email-password#update-password

I think we'll need to use `/set-password` on the server: https://better-auth.com/docs/concepts/users-accounts#set-password

What do you think? @nahSystemu

nahSystemu:
> I'm not sure this approach will work:
> 
> https://better-auth.com/docs/authentication/email-password#update-password
> 
> Looks like previous password is a requirement of `/change-password`: https://better-auth.com/docs/authentication/email-password#update-password
> 
> I think we'll need to use `/set-password` on the server: https://better-auth.com/docs/concepts/users-accounts#set-password
> 
> What do you think? @nahSystemu

Didn't notice that when updating, here is it using the setPassword from server

hjball (approved):
Nice one @nahSystemu

PR #486, opened 2026-05-02
feat: websocket integration

### LINKED ISSUE #7: Add real-time

nahSystemu:
## Summary

Closes #7 

---

This PR adds a WebSocket server to Kan and wires it up across the API and frontend to enable real-time updates for everyone on a workspace.

The server sits as its own service alongside the API and handles event delivery. When a mutation happens — a card is moved, a comment is added, a label changes, anything — the API publishes an event to the WebSocket server, which fans it out to all connected clients watching that workspace. On the frontend, the relevant queries are invalidated and the UI updates automatically. To avoid unnecessary re-fetches, each event includes the ID of the user who triggered it so clients can skip updates they caused themselves. Subscriptions are also scoped to the active board so clients aren't processing events that have nothing to do with what they're looking at.

A few things worth noting about the implementation:

- The server uses Redis pub/sub when available so it works in multi-instance deployments, with an in-process EventEmitter as a fallback for simpler setups.
- If a client goes offline, mutations are queued locally and replayed on reconnect, and all queries are refreshed so nothing is left stale.
- Events are only emitted after the mutation has fully committed to the database, so clients never receive partial or inconsistent state.

Alongside the core event system, this also adds a presence indicator in the board toolbar (live viewer avatars), an in-app notification inbox for things like card assignments and comments, and rate limiting on high-frequency card mutations. Webhook coverage was also expanded to include boards, lists, labels, and comments while the infrastructure was being touched.

## How did you test this change?

Tested locally via Docker Compose with the full stack running. Verified live updates across multiple browser tabs for all major mutation types, presence updates, and notification delivery. Offline queuing was tested by disabling the network mid-session and confirming mutations replayed correctly on reconnect.


nahSystemu:
@hjball Handing this off to you, needs some updating prior to merging, spent an hour already fixing merge conflicts and thought I was on latest, but can see that its 15 commits behind, so good luck. Test it on your end, see how it works, if anything is off tell me, I'll remove or change what needs to be changed

PR #490, opened 2026-05-10
Add cardPublicId to card webhook payloads

Midas-sudo:
I would like to suggest the inclusion of the card public id on the fired events of the webhooks.

I would like to partially automate label tagging and for that I would require to know the public id of the card in order to manipulate it.


hjball (approved):
lgtm @Midas-sudo - ideally we should only be exposing the publicId but we can address that separately

Midas-sudo:
I assumed as such but since removing the id could break possible automations of other people I decided to only propose the addition of the publicCardId.

PR #495, opened 2026-05-13
fix(s3): default S3_REGION to us-east-1 instead of empty string

nickmeinhold:
## Summary

The AWS SDK throws `Region is missing` when `S3Client` is constructed with an empty-string region, so any S3 feature (avatars, attachments, presigned URL generation) is silently broken when `S3_REGION` is unset — even though `.env.example` ships it unset and `apps/web/src/env.ts` declares it `optional()`. The schema and the runtime disagreed about whether the var was required.

This PR defaults to `us-east-1` in `createS3Client`. S3-compatible providers (MinIO, Backblaze B2, R2, DigitalOcean Spaces, Wasabi) ignore the region entirely; real AWS S3 users should set `S3_REGION` explicitly to their bucket's actual region (clarified in `.env.example`).

Found while debugging an unrelated PDF-attachment issue (kanbn/kan#494) — couldn't reproduce locally because the smoke harness exploded on `Region is missing` before reaching the actual code path.

## Test plan
- [ ] With `S3_REGION` unset, attachment upload no longer 500s on `S3Client` construction
- [ ] With `S3_REGION` explicitly set (e.g. `eu-west-1`), behaviour is unchanged
- [ ] MinIO / R2 / Spaces deployments unaffected

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,8 +23,8 @@ SMTP_REJECT_UNAUTHORIZED= # set to "false" to accept invalid certs
 NEXT_PUBLIC_DISABLE_EMAIL=
 
 # S3 storage (optional)
-S3_REGION=
-S3_ENDPOINT=
+S3_REGION= # required for AWS S3 (e.g. us-east-1); ignored by most S3-compatible providers — defaults to us-east-1 if unset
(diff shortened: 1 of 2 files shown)


hjball (approved):
Seems sensible to me @nickmeinhold

PR #510, opened 2026-06-02
feat: add mobile list snap scrolling

knom:
Adding snapping logic, so that it snaps to columns while horizonally scrolling. 

Note: It doesn't work for the last column.

### DIFF
--- a/apps/web/src/views/board/components/List.tsx
+++ b/apps/web/src/views/board/components/List.tsx
@@ -95,7 +95,7 @@ export default function List({
           ref={provided.innerRef}
           {...provided.draggableProps}
           {...provided.dragHandleProps}
-          className="dark-text-dark-1000 mr-5 h-fit min-w-[18rem] max-w-[18rem] rounded-md border border-light-400 bg-light-300 py-2 pl-2 pr-1 text-neutral-900 dark:border-dark-300 dark:bg-dark-100"
+          className="dark-text-dark-1000 mr-5 h-fit min-w-[18rem] max-w-[18rem] snap-start rounded-md border border-light-400 bg-light-300 py-2 pl-2 pr-1 text-neutral-900 md:snap-none dark:border-dark-300 dark:bg-dark-100"
         >
           <div className="mb-2 flex justify-between">
             <form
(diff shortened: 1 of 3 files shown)


knom:
@nickmeinhold are you the one to approve this one?

nickmeinhold:
@knom i don't have merge privileges but can review. I think I might be able to give a review approval although I think it is only signal and @hjball has the final call

nickmeinhold (approved):
Approving as signal (note: I don't have merge rights, so @hjball has the final call).

Nice, well-scoped change. Snapping feels good on mobile and the `md:snap-none` keeps desktop untouched. A few notes, none blocking the approval but worth tracking:

**Follow-up to track (your own caveat):** the last column being unreachable is a real `scroll-snap-type: mandatory` quirk: the last list has nothing to scroll past, so it can't align to `start` and gets pulled back. A few ways to fix when you get to it:
- switch `snap-mandatory` to `snap-proximity` (gentlest, lets the tail rest anywhere), or
- add `snap-align: end` to the last list, or
- add a trailing spacer wide enough for the last list to reach `start`.

**Minor:** the leading spacer change `min-w-[2rem]` to `min-w-[10px]` applies at all breakpoints, not just mobile, so it tightens the desktop left gutter from 32px to 10px. If that desktop change isn't intended, you may want to guard it behind a breakpoint.

**Trivial:** `md:snap-none` on the list item in `List.tsx` is a no-op. `snap-none` sets `scroll-snap-type`, which is a container property, so it has no effect on an item. The real desktop-disable on the parent container is correct.

**Worth a quick device test:** since the board uses drag-and-drop, it's worth confirming `snap-mandatory` doesn't fight horizontal touch-drag / auto-scroll on a real phone.

nickmeinhold:
@knom approved as signal. The only thing I'd actually flag before this lands is the last-column issue you noted: switching `snap-mandatory` to `snap-proximity` (or adding `snap-align: end` / a trailing spacer) should make the last list reachable. The two other notes in my review (the desktop left-gutter change from the spacer, and a no-op `md:snap-none` on the list item) are minor and your call.

I can't merge it myself though, so @hjball has the final say on landing it.

hjball (approved):
lgtm, nice UX improvement @knom

Issue #155, opened 2025-08-19
Deeper theme customization with colors and pictures


TheApache142857:
## ✨ Feature Request

**Is your feature request related to a problem? Please describe.**
Compared to other solutions, I don't have a much visibility with colors and pictures

**Describe the solution you'd like**
- Being able to add a color or picture bar to the top of cards
- Add a custom background

**Additional context**
Here is what my board looks like on another solution
<img width="1417" height="1365" alt="Image" src="https://github.com/user-attachments/assets/0b058805-d1ad-4273-b756-240d89e9e0d3" />
Compared to an all white board I admit it's a bit more intriguing to work on

**Would you like to work on this feature?**

- [X] Yes, I'd like to help implement this feature (But I'd take me quite some time)
- [ ] No, I'm just suggesting the feature

Eshwar1212-maker:
@TheApache142857 Hey, would you want the solution to have an option to upload background colors with just CSS on the board? Or actual full pictures? If for pull pictures that would require a bigger solution with s3 or some cloud storage that a few of the other tickets work on. 

If you mean just with random colors users can choose then that would be easiar and id like to work on it.

TheApache142857:
@Eshwar1212-maker Hi! Yeah Trello has sort of a pallet you can choose colors from, that would be a fantastic start.

<img width="750" height="1710" alt="Image" src="https://github.com/user-attachments/assets/cfed0f0f-f50e-4e5e-8820-48f0faa0d464" />

For full pictures we can wait for cloud/local storage to be worked on.

nickmeinhold:
Hey @hjball — we'd love to take a crack at this if it's something you'd accept a PR for. Wanted to check before investing the effort.

Looking at the codebase, the infrastructure is already there to support this nicely:
- S3 presigned URL flow for uploads (avatars + card attachments already use it)
- The existing `colours` palette in `@kan/shared/constants` for a solid-color picker
- `PatternedBackground` component that could be extended for custom board backgrounds

We'd propose a phased approach:

**Phase 1 — Card cover colors**
Add a `coverColourCode` field to cards. Users pick from the existing palette (or custom hex). Renders as a colored bar at the top of the card component. Lightweight, no upload infra needed.

**Phase 2 — Card cover images**
Add `coverImageS3Key` to cards, reusing the attachment upload pattern. Image renders as the card cover, resized/cropped to fit.

**Phase 3 — Board backgrounds**
[...]

PR #513, opened 2026-06-04
feat: Board background settings

knom:
A first draft of a background option for boards.

Adds a settings menu on board (lookout for the three dots).
And let's you pick a background there.

First draft as I said.. The picker UX should probably be better.. But looking to find out if this is the right direction for folks..

Thanks.


nickmeinhold (commented):
Really nice first slice @knom — keeping it scoped to just the board background is the right call, and a couple of things stood out as genuinely well done:

- The `hasOtherUpdates` change from `input.name || …` to `input.name !== undefined` quietly fixes a latent bug (an empty-string name was previously falsy). Good catch.
- The luminance-based contrast for the empty-state title is a thoughtful touch.

A few suggestions, roughly in priority order:

**1. Palette over the raw colour input (the main one).** The issue (#155) is really asking for a Trello-style swatch palette, and you flagged the picker UX yourself. There's already a shared palette at `@kan/shared/constants` (`colours`) that labels use — offering those swatches as the primary picker (with the native `<input type="color">` as an optional "custom" escape hatch) would match the request and stay consistent with the rest of the app. For reference, I just used that palette for card cover colours in #515 if a concrete example helps — and the two PRs are nicely complementary (board background vs card cover, no overlap).

**2. Contrast handling on the public board.** `views/public/board/index.tsx` renders the coloured background but doesn't apply the contrast-adjusted text class, so light backgrounds could make text hard to read there. Relatedly, `isLightHexColor` is inlined in `board/index.tsx` — might be worth lifting into a small shared util so both views (and the board header text, not just the empty state) can use it.

**3. `INITIAL_BOARD_BACKGROUND_COLOR = "#0d9488"`** is exactly the palette's "Teal" — importing it from `colours` instead of hardcoding the hex would avoid the magic literal.

**4. Minor:** the `useEffect` that focuses `#board-background-color` still references `nameElement` (copy-paste from a name form), and auto-focusing a colour input is a slightly unusual UX — probably safe to drop.

Happy to help on any of these. Great direction overall.

knom:
> Really nice first slice @knom — keeping it scoped to just the board background is the right call, and a couple of things stood out as genuinely well done:
> 
> * The `hasOtherUpdates` change from `input.name || …` to `input.name !== undefined` quietly fixes a latent bug (an empty-string name was previously falsy). Good catch.
> * The luminance-based contrast for the empty-state title is a thoughtful touch.
> 
> A few suggestions, roughly in priority order:
> 
> **1. Palette over the raw colour input (the main one).** The issue (#155) is really asking for a Trello-style swatch palette, and you flagged the picker UX yourself. There's already a shared palette at `@kan/shared/constants` (`colours`) that labels use — offering those swatches as the primary picker (with the native `<input type="color">` as an optional "custom" escape hatch) would match the request and stay consistent with the rest of the app. For reference, I just used that palette for card cover colours in #515 if a concrete example helps — and the two PRs are nicely complementary (board background vs card cover, no overlap).
> 
> **2. Contrast handling on the public board.** `views/public/board/index.tsx` renders the coloured background but doesn't apply the contrast-adjusted text class, so light backgrounds could make text hard to read there. Relatedly, `isLightHexColor` is inlined in `board/index.tsx` — might be worth lifting into a small shared util so both views (and the board header text, not just the empty state) can use it.
> 
> **3. `INITIAL_BOARD_BACKGROUND_COLOR = "#0d9488"`** is exactly the palette's "Teal" — importing it from `colours` instead of hardcoding the hex would avoid the magic literal.
> 
> **4. Minor:** the `useEffect` that focuses `#board-background-color` still references `nameElement` (copy-paste from a name form), and auto-focusing a colour input is a slightly unusual UX — probably safe to drop.
> 
> Happy to help on any of these. Great direction overall.

Great feedback, thank you!

I've started to incorporate it and have just pushed changes for #2.
The rest is to follow!

thx.

knom:
Okay.. Here's my V1
Text color stays static for now, as I was messing too much with the CSS classes and light/dark system theme.. :-(

Looking forward to advixe on how to do the text color switching best!

@nickmeinhold @hjball please review and let me know what you think!

scoobs-mystery-bytes:
Wondering if in addition to allowing colors for the board, to allow for a background image to take it a step further?

knom:
> Wondering if in addition to allowing colors for the board, to allow for a background image to take it a step further?

Absolutely!
This is a natural step 2.

But I wanna see with @hjball first what he thinks of the overall approach!

PR #514, opened 2026-06-04
fix: correct German gender for "Neues Board"

hajorappe:
"Board" is neuter in German, so "Neue Board" should be "Neues Board".


hjball (approved):
Good spot

PR #515, opened 2026-06-05
feat: card cover colours (Phase 1 of #155)

nickmeinhold:
## Card cover colours (Phase 1 of #155)

Adds an optional **cover colour** to cards, rendered as a bar across the top of the card on both private and public boards. Set it from the card context menu → **Set cover colour** (palette swatches from `@kan/shared`), or clear back to none.

This is the lightweight, palette-based first phase of the theme-customisation request in #155 — no upload infrastructure needed. It deliberately **complements @knom's #513** (board backgrounds): different field, different surface, no overlap. (Coordination on the issue: https://github.com/kanbn/kan/issues/155#issuecomment-4629076532)

### What's in it
- **db**: nullable `card.coverColourCode` (`varchar(7)`) + generated migration
- **api**: `coverColourCode` threaded through `card.update` input → repo → `cardDetail`/`boardDetailCard` output schemas → board query selection
- **web**: cover bar on `Card`; new `CardContextCoverColourModal` (palette + clear) wired into the card context menu; optimistic update via the existing `invalidateCard` pattern

### Design notes
- Stores the palette hex (`#0d9488`…) for consistency with how labels store `colourCode`.
- Bidirectional-safe: `null` clears, `undefined` leaves untouched (mirrors the existing `dueDate` handling in `card.update`).
- Validation (`/^#[0-9a-fA-F]{6}$/`) lives at the router boundary.

### Verification
- ✅ `@kan/db`, `@kan/api` typecheck clean; `@kan/web` introduces **zero** new type errors vs `main` (diffed against baseline).
- ⏳ **Not yet runtime-verified** by me against a live instance — opening as **draft** for exactly that reason. Happy to add a screenshot / screen-recording and any UX adjustments. Feedback on the picker UX welcome.

Closes part of #155 (Phase 1; cover **images** and board-level options tracked separately).

### DIFF
--- a/apps/web/src/views/board/components/Card.tsx
+++ b/apps/web/src/views/board/components/Card.tsx
@@ -24,9 +24,11 @@ const Card = ({
   comments,
   attachments,
   dueDate,
+  coverColourCode,
 }: {
   title: string;
   ticketNumber?: string | null;
+  coverColourCode?: string | null;
   labels: { name: string; colourCode: string | null }[];
   members: {
     publicId: string;
@@ -69,6 +71,13 @@ const Card = ({
 
   return (
     <div className="flex flex-col overflow-hidden rounded-md border border-light-200 bg-light-50 px-3 py-2 text-sm text-neutral-900 dark:border-dark-200 dark:bg-dark-200 dark:text-dark-1000 dark:hover:bg-dark-300">
+      {coverColourCode && (
+        <div
+          className="-mx-3 -mt-2 mb-2 h-1.5"
+          style={{ backgroundColor: coverColourCode }}
+          aria-hidden="true"
+        />
+      )}
       {ticketNumber && (
         <span className="mb-1 text-xs text-light-700 dark:text-dark-800">
           {ticketNumber}
--- a/apps/web/src/views/board/components/CardContextCoverColourModal.tsx
+++ b/apps/web/src/views/board/components/CardContextCoverColourModal.tsx
@@ -1,0 +1,110 @@
+import { t } from "@lingui/core/macro";
+import { HiCheck, HiXMark } from "react-icons/hi2";
+
+import { colours } from "@kan/shared/constants";
+
+import { useModal } from "~/providers/modal";
+import { usePopup } from "~/providers/popup";
+import { api } from "~/utils/api";
+import { invalidateCard } from "~/utils/cardInvalidation";
+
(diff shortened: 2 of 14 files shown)


nickmeinhold:
Verified locally against a running instance (Next dev + Postgres):

- Render path: cards with a `coverColourCode` show the colour bar across the top on the board (seeded a teal card directly in the DB → rendered correctly).
- Setter path: right-click card → **Set cover colour** → picked Orange from the palette → the `card.update` mutation persisted `#ea580c` and the bar re-rendered immediately.
- Picker is the palette-swatch UX from `@kan/shared` (clear/none option + the 8 shared colours).

Screenshots to follow. Will flip out of draft once they're attached.

knom:
Hey @nickmeinhold,
I love the work you're doing here! 

Why not move to the general of concept of a settings menu and a popup dialog. 
Like I also did for board, same would make sense for card!

A central place for item related configuration like also color..

What's your thoughts on that?

nickmeinhold:
Screenshots from the local verification run:

kan-515-cards-with-cover-colours
kan-515-cover-colour-picker

nickmeinhold:
Thanks @knom, really appreciate the enthusiasm, and I think you're onto something for the longer term.

Before either of us invests more, I think this is worth a maintainer steer so we build to the project's convention.

Cards already have two config surfaces today: the per-action context-menu items (Manage members, Add / edit label, Set due date), each opening a focused modal, plus the full card detail view where description, labels, members and due date are edited. So unlike boards (which had no config hub, which is exactly why your board settings dialog made sense there), cards already have a "central place".

That makes the cover-colour placement a genuine convention question:

- **A:** keep it as a focused context action alongside the existing card-context items (what #515 does now)
- **B:** a unified card settings dialog mirroring board settings, with colour as one section
- **C:** fold card visual settings into the existing card detail view rather than a new dialog

@hjball, do you have a preference for how card-level configuration should be organised? Happy to shape #515 to match whatever direction you want, and it'd keep knom's #513 and this PR consistent with each other.

nickmeinhold:
And @knom, separately from all the design back-and-forth: it's genuinely great to be collaborating on this with you. Two complementary PRs landing side by side, feedback flowing both ways, this is open source at its best. Looking forward to seeing where it goes.

knom:
Same here!
This is fun 😉

PR #523, opened 2026-06-19
fix: sync Editor editable state with readOnly prop

### LINKED ISSUE #522: Card description Tiptap editor stuck permanently read-only
**Source**: `apps/web/src/components/Editor.tsx`
**Root cause**: `useEditor` empty dependency array + race condition on `readOnly` prop
**Affects**: Any user where permissions query resolves after the card query

### `Editor.tsx` line 460-578

```tsx
const editor = useEditor(
  {
    // ...
    editable: !readOnly,       // captured once at creation time
    onUpdate: ({ editor }) => onChange?.(editor.getHTML()),
    onBlur: ({ event }) => { /* ... */ },
  },
  [],  // <-- empty array: editor created ONCE, never re-configured
);
```

There is no `useEffect` to sync `editor.setEditable(!readOnly)` when the `readOnly` prop changes after mount. The empty dependency array `[]` means the editor configuration is frozen at creation time.

### Race condition in `CardPage` (`apps/web/src/views/card/index.tsx`)

```tsx
const { canEditCard } = usePermissions();       // async API query
const { data: card } = api.card.byId.useQuery(…);
const isCreator = card?.createdBy && session?.user.id === card.createdBy;
const canEdit = canEditCard || isCreator;        // false on first render

// Editor only renders when card is loaded
{card && (
  <Editor
    readOnly={!canEdit}     // true on first render
    onChange={canEdit ? (e) => setValue("description", e) : undefined}
    onBlur={canEdit ? () => handleSubmit(onSubmit)() : undefined}
  />
)}
```

**Timeline:**

| Step | Card query | Permissions query | `canEdit` | Editor state |
|------|-----------|-------------------|-----------|-------------|
| 1 | loading | loading | `false` | not rendered |
| 2 | **resolved** | still loading | `false` | **renders with `readOnly=true`** |
| 3 | cached | **resolved** | `true` | prop changes but `useEditor` never updates |

Because the editor is created inside `{card && (...)`, it only appears once the card query resolves. If the permissions query is still in-flight at that point (which it usually is — they're independent), the editor is created with `editable: false`, `onChange: undefined`, `onBlur: undefined` — permanently.

### Symptoms

- Description field appears as plain text, not editable
- No visual indication it can be edited (no placeholder text, no cursor change)
- Clicking into the description area does nothing
- Works correctly only if the permissions query resolves *before* the card query (unreliable — depends on cache state, network timing)

## The Fix

Add a `useEffect` in `Editor.tsx` to sync `editable` state when `readOnly` prop changes:

```tsx
// After the useEditor call and before the content-sync useEffect
useEffect(() => {
  if (!editor) return;
  if (editor.isEditable === !readOnly) return;
  editor.setEditable(!readOnly);
}, [editor, readOnly]);
```

For the `onChange`/`onBlur` closure capture issue (same root cause — empty dep array), the handlers also need to be updated when they change. Either:

1. **Re-create the editor** when callbacks change (add deps to `useEditor`), or
2. **Use refs** for the callbacks (common Tiptap pattern):

```tsx
const onChangeRef = useRef(onChange);
onChangeRef.current = onChange;
// In useEditor config:
onUpdate: ({ editor }) => onChangeRef.current?.(editor.getHTML()),
```

## Also Affects

- Title textarea in CardPage uses `register("title")` directly from react-hook-form (not Tiptap), so it does NOT have this bug. Only the description field is affected.
- The `Placeholder` plugin config also captures the initial `readOnly` value at line 478-482 (shows empty placeholder when `readOnly=true`, never updates).
- The `SlashCommands` and `Mention` plugin configs are similarly frozen.

## Smoking Gun: PlainTextEditor already has the fix

The maintainers already know about this pattern — they applied the exact fix to `apps/web/src/components/PlainTextEditor.tsx` (lines 28-48, 105-108):

```tsx
// refs for all callbacks
const onEnterRef = useRef(onEnter);
const onEscapeRef = useRef(onEscape);
const onBlurRef = useRef(onBlur);
const onChangeRef = useRef(onChange);

useEffect(() => { onChangeRef.current = onChange; }, [onChange]);
// ... etc

// and the setEditable sync
useEffect(() => {
  if (!editor) return;
  editor.setEditable(!readOnly);
}, [readOnly, editor]);
```

The same fix was simply never applied to `Editor.tsx`. This makes the bug an oversight, not a design choice — lowering the bar for upstream acceptance.

(Note: `Comment.tsx` sidesteps the bug entirely by using two separate `<Editor>` instances behind an `isEditing` ternary, so `readOnly` never mutates on a live instance. Only `CardPage` mounts a single Editor whose `readOnly` prop flips from `true` → `false`.)

## Verified Against

- Upstream HEAD: `f369ebe` (tag `v0.5.6`) — bug is present
- `@tiptap/react@^2.14.0` — `editor.setEditable()` is the documented API
- `Editor.tsx` typechecks clean after the fix (zero new TS errors)
- `git diff --stat`: 1 file changed, 23 insertions(+), 2 deletions(-)

## Rediscovery

This bug was noticed by a user while testing the kan-zitadel deployment. The initial suspicion was a permission/role issue (guest role lacking `card:edit`), but the user was an admin with full permissions. The actual cause is the Tiptap `useEditor` initialization timing.

JayDataEngineer:
## Description

The rich-text `Editor` component (used for card descriptions) creates its Tiptap instance once via `useEditor` with an empty dependency array. This means:

1. `editable: !readOnly` is captured at mount time and never updated.
2. `onChange` / `onBlur` are frozen as the first-render closures.

In `CardPage`, the description `Editor` mounts as soon as the card query resolves — but the permissions query is usually still in-flight at that point. So on first render `readOnly={true}` and the handlers are `undefined`. When permissions resolve a moment later and `canEdit` flips to `true`, the `Editor` never becomes editable — leaving the description permanently stuck read-only, even for workspace admins.

Refreshing the page sometimes "fixes" it because cached permission state resolves faster on subsequent loads. This explains why the issue is intermittent and hard to reproduce on a fast connection with a warm cache.

### Why this is a real oversight

The existing `PlainTextEditor.tsx` already uses the exact pattern this PR applies:
- Refs for callbacks (`onChangeRef`, `onBlurRef`, etc.)
- `useEffect` syncing `editor.setEditable(!readOnly)` when the prop changes

It just never got ported to `Editor.tsx`. `Comment.tsx` sidesteps the issue by mounting two separate `<Editor>` instances behind an `isEditing` ternary, so `readOnly` never mutates on a live instance — but `CardPage` mounts a single `Editor` whose `readOnly` prop flips from `true` → `false`, which is where this bites.

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

- [x] I have included screenshots for any UI changes (no UI changes)

## Verification

- `Editor.tsx` typechecks clean (no new TS errors vs. HEAD)
- Diff is minimal: 1 file, +23 / -2
- Fix mirrors the proven pattern in `PlainTextEditor.tsx`

## Linked issue

Closes https://github.com/kanbn/kan/issues/522

## Notes for reviewer

- Did not re-run prettier on the whole file (it would have produced ~80 lines of unrelated formatting noise from pre-existing issues in `Editor.tsx`). Only the fix lines are touched.
- The `Placeholder` plugin config also captures `readOnly` at creation time and won't show its hint text after `setEditable(true)`. Not addressed here — it's cosmetic and would require extension reconfiguration. Happy to follow up if desired.

### DIFF
--- a/apps/web/src/components/Editor.tsx
+++ b/apps/web/src/components/Editor.tsx
@@ -457,6 +457,17 @@ export default function Editor({
 }) {
   const containerRef = useRef<HTMLDivElement>(null);
 
+  // useEditor is created once (empty deps below), so keep the latest callbacks
+  // in refs to avoid the editor capturing stale closures on re-render.
+  const onChangeRef = useRef(onChange);
+  const onBlurRef = useRef(onBlur);
+  useEffect(() => {
+    onChangeRef.current = onChange;
+  }, [onChange]);
+  useEffect(() => {
+    onBlurRef.current = onBlur;
+  }, [onBlur]);
+
   const editor = useEditor(
     {
       extensions: [
@@ -553,7 +564,7 @@ export default function Editor({
         ...(enableYouTubeEmbed ? [YouTubeNode] : []),
       ],
       content,
-      onUpdate: ({ editor }) => onChange?.(editor.getHTML()),
+      onUpdate: ({ editor }) => onChangeRef.current?.(editor.getHTML()),
       onBlur: ({ event }) => {
         if (
           document
@@ -563,7 +574,7 @@ export default function Editor({
           return;
         // Only trigger onBlur if the click was outside both the editor and menu
         if (!containerRef.current?.contains(event.relatedTarget as Node)) {
-          onBlur?.();
+          onBlurRef.current?.();
         }
       },
       editorProps: {
@@ -586,6 +597,16 @@ export default function Editor({
       editor.commands.setContent(safeContent, false);
     }
   }, [content, editor]);
+
+  // useEditor captures `readOnly` once at creation time (empty deps above), so
+  // explicitly sync `editable` when the prop changes. Without this the editor
+  // gets stuck read-only when `readOnly` flips from true to false after mount
+  // (e.g. card permissions resolving slower than the card data on first load).
+  useEffect(() => {
+    if (!editor) return;
+    if (editor.isEditable === !readOnly) return;
+    editor.setEditable(!readOnly);
+  }, [editor, readOnly]);
 
   return (
     <div ref={containerRef}>


JayDataEngineer:
Closes #522

hjball (approved):
lgtm, nice catch @JayDataEngineer

PR #545, opened 2026-07-31
fix: accept empty slug in workspace.create

HeinrichNebula:
## Problem
`NewWorkspaceForm` submits `slug: ""` when the URL field is left blank — its client-side schema explicitly allows this (`.optional().or(z.literal("")))`. The server-side input schema of `workspace.create` however only allows `undefined` or a valid 3–64 char slug, so the mutation always fails validation with `too_small` + regex errors on `path: ["slug"]`.

On self-hosted instances this means **workspace creation is broken whenever the URL field is left empty** (reproducible on current main; observed on a self-hosted instance pinned to the latest ghcr image).

## Fix
- Mirror the client schema on the server: `.optional().or(z.literal(""))`.
- Fall back to the generated public id with `input.slug || workspacePublicId` (`??` keeps the empty string).

The existing `if (input.slug)` guards (cloud check, availability/reserved checks) already treat `""` as absent, so no further changes are needed.

### DIFF
--- a/packages/api/src/routers/workspace.ts
+++ b/packages/api/src/routers/workspace.ts
@@ -205,12 +205,18 @@ export const workspaceRouter = createTRPCRouter({
       z.object({
         name: z.string().min(1).max(64),
         description: z.string().max(280).optional(),
+        // NewWorkspaceForm submits slug as "" when the user leaves the URL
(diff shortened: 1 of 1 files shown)


hjball (approved):
Good catch @HeinrichNebula

hjball (approved):

PR #548, opened 2026-08-06
feat(card): attach files directly when creating a card

HeinrichNebula:
## What

The new-card modal only supports title/description/labels/members/due date. Attaching a file requires creating the card first, reopening it, and uploading there. This PR lets users queue files in the create form itself:

- Paperclip button (hidden multi-file input) plus drag & drop onto the form
- Queued files render as removable chips below the option row
- On submit, the card is created first, then the files are uploaded sequentially to the existing `/api/upload/attachment` endpoint, followed by a board invalidation
- If the card mutation fails, the queue is restored so a retry keeps the files
- If some uploads fail, the card is kept and a popup reports how many attachments failed (no rollback)
- "Create another" clears the queue

Also removes duplicate `comments`/`checklists`/`attachments` keys in the optimistic card payload in the same file (TS1117, a hard error under `tsc --noEmit`).

## Notes

- Files are kept in component state, not in the persisted modal form state (File objects do not survive serialization)
- No new endpoints or schema changes; reuses the upload path and its 50MB limit
- i18n via the existing lingui `t` macros

## Testing

- Create card with 2 files: both appear as attachments on the created card
- Create card without files: behavior unchanged
- Upload failure (oversized file): card exists, popup reports the failed count
- "Create another": queue resets between cards

### DIFF
--- a/apps/web/src/views/board/components/NewCardForm.tsx
+++ b/apps/web/src/views/board/components/NewCardForm.tsx
@@ -1,8 +1,10 @@
 import { t } from "@lingui/core/macro";
 import { Trans } from "@lingui/react/macro";
 import { format } from "date-fns";
-import { useEffect, useState } from "react";
+import { env } from "next-runtime-env";
+import { useEffect, useRef, useState } from "react";
 import { useForm } from "react-hook-form";
+import { HiOutlinePaperClip } from "react-icons/hi";
 import {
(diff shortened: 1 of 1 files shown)


hjball (approved):
Please raise an issue and wait for approval before opening a feature PR in future @HeinrichNebula 

This particular change seems sensible so happy to merge this as a one off

PR #555, opened 2026-08-21
perf: improvements to build time

nahSystemu:
The Docker workflow was slow for two reasons, and both of them are easy to miss unless you read the build log properly. The arm64 half of the web image was getting built under QEMU emulation, and the layer cache never hit anything, so every run rebuilt from scratch.

The arm64 Next build went from `14m01.956s` to `1m36.583s` on the same machine, same output, still two multi-arch images signed with cosign.

---

**What was actually wrong**

- BuildKit only advertises x86 platforms, so every arm64 stage went through the host binfmt handlers. `pnpm install` was 5.5x slower, `turbo run build` 6.2x slower.
- Both build steps wrote to the same gha cache scope, so each one overwrote the other's manifest. A cache got imported every run and then everything rebuilt anyway, zero `CACHED` steps.
- The pnpm store mount pointed at `~/.pnpm-store`. BuildKit doesn't expand `~` in mount targets but pnpm does, so the two never pointed at the same directory and the mount did nothing. Every run logged `resolved 0, reused 0, downloaded 111`.

---

**The fix**

The base stage is pinned to `$BUILDPLATFORM` now, so `pruner`, `deps` and `builder` all run native no matter the target arch. Only the final distroless stage is per arch and that one just copies plain JS, which is already how the image worked, it builds on musl and runs on glibc.

`migrate` is left per arch on purpose. drizzle-kit pulls esbuild, which has a per platform binary, so cross compiling it would drop `@esbuild/linux-x64` into the arm64 image. It stays emulated at about 71s, which is worth it.

Everything else is cache and cleanup:

- pnpm store mount fixed with an absolute path, plus new cache mounts for `.next/cache`, `.turbo` and npm.
- gha cache split into `scope=web` and `scope=migrate` so the two builds stop fighting over the same key.
- QEMU action gets called explicitly now. It only worked before cause the runner happened to have binfmt set up, and `migrate` still needs arm64 for real.
- Actions bumped to current SHA pinned releases, which drops the `cosign-release: "v2.2.4"` pin since v4 picks its own default.
- Git version logic taken out of `pruner`. `.dockerignore` excludes `.git`, so every git call was already doing nothing and the version comes in as a build arg anyway. `fetch-tags: true` added on checkout, without it the version falls back to a bare sha and the version link in the user menu points at a release tag that doesn't exist.
- Only PR builds get cancelled by the concurrency group now, branch and tag builds publish things so they get to finish.

---

**The trade-off**

Cross compiling broke sharp, which is what the `supportedArchitectures` change in `package.json` is for. sharp picks a prebuilt binary per platform, so building `deps` on amd64 put x64 binaries into the arm64 image, and Next has image optimization on, so that would have died at runtime. The emulated build only got it right by accident.

Installing every Linux variant fixes it but adds about 51MB to the image. If that's too much, the other option is a small per arch stage that installs only sharp, but the Dockerfile gets messier. Any future dep with a native addon needs the same treatment, that's just what you get with cross compiling.

---

**Checked**

`docker build --check` is clean on both targets. The arm64 image reports `linux/arm64` and sharp loads in it, the migrate image has `@esbuild/linux-arm64`, and `pnpm install --frozen-lockfile` still passes with the lockfile unchanged.

Timings were on an 8 core machine and `ubuntu-latest` has 4, so the absolute numbers will differ but the ratios shouldn't. Splitting web and migrate into parallel jobs would save another couple of minutes, but that's not in this PR.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -30,7 +30,9 @@ jobs:
   build:
     concurrency:
       group: docker-build-${{ github.ref }}
-      cancel-in-progress: true
+      # Only supersede PR builds. Branch and tag builds publish artifacts and
+      # are allowed to finish.
+      cancel-in-progress: ${{ github.event_name == 'pull_request' }}
     runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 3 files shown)


hjball (approved):
Nice one @nahSystemu

PR #561, opened 2026-08-27
fix: create empty boards through MCP

mikkelvalentinsorensen:
## Summary
- supply empty `lists` and `labels` arrays required by the Kan board-create API while preserving the minimal MCP tool input
- add an MCP tool contract test for the forwarded payload

Cross-references https://github.com/mikkelvalentinsorensen/MikValSor.HomeLab.Docker01/issues/259.


hjball (approved):
Nice one @mikkelvalentinsorensen

PR #563, opened 2026-08-27
Fix: side navigation labels do not update on locale change

serhiihordiichuk:
## Problem

The side navigation labels (Boards, Templates, Members, Settings) stay in the locale that was active at first render. When a user switches the language, the rest of the interface updates, but the sidebar does not. A hard reload does not help either, because the saved locale activates after the first render.

## Cause

`SideNavigation.tsx` builds the navigation items inside a `useMemo` with `[isDarkMode]` as its only dependency. The `t` macro values are computed once and cached, so a locale change never recomputes them.

## Fix

Read the active locale from the existing `useLocalisation` hook and add it to the dependency array. The memo now recomputes when the language changes.

This affects every language, not one specific locale. Related to #562 (where the issue was found while testing the Ukrainian translation).

### DIFF
--- a/apps/web/src/components/SideNavigation.tsx
+++ b/apps/web/src/components/SideNavigation.tsx
@@ -16,6 +16,7 @@ import type { Subscription } from "@kan/shared/utils";
 import { hasActiveSubscription } from "@kan/shared/utils";
 
 import type { KeyboardShortcut } from "~/providers/keyboard-shortcuts";
+import { useLocalisation } from "~/hooks/useLocalisation";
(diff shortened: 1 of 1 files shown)


hjball (commented):
Good catch @serhiihordiichuk - worth fixing in the other places where we haven't included i18n.locale in the deps

serhiihordiichuk:
> Good catch @serhiihordiichuk - worth fixing in the other places where we haven't included i18n.locale in the deps

Let me know if you would like to include other places in this PR

hjball:
Please @serhiihordiichuk

PR #564, opened 2026-08-27
Fix: boards page heading and tabs are not translated

serhiihordiichuk:
## Problem

On the boards page, the "Active" / "Archived" tabs and the "Boards" / "Templates" heading always render in English, even when the user selects another language. Translations for all four strings exist in the catalogs.

## Cause

Two separate issues in `views/boards/index.tsx`:

1. The tabs array was a module-level constant. The `t` macro values were computed once at module load, before any locale activates.
2. The heading used `t`${isTemplate ? "Templates" : "Boards"}`` — the ternary passes a raw English string into a placeholder, so the catalog translation never applies.

## Fix

- Move the tabs array into the component and subscribe with `useLingui()`, so the labels re-render after locale activation.
- Translate the heading words directly: `isTemplate ? t`Templates` : t`Boards``. The page title uses the same approach.

No catalog changes are needed — all message ids already exist.

Same bug class as #563, found while testing the Ukrainian translation from #562.

### DIFF
--- a/apps/web/src/views/boards/index.tsx
+++ b/apps/web/src/views/boards/index.tsx
@@ -5,6 +5,7 @@ import {
   ListboxOptions,
 } from "@headlessui/react";
 import { t } from "@lingui/core/macro";
+import { useLingui } from "@lingui/react";
 import { useMemo, useState } from "react";
 import {
   HiArrowDownTray,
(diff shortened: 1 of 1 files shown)


hjball (approved):
lgtm

PR #567, opened 2026-09-01
fix: use public base URL for login redirects

bionicman:
## Description

In a self-hosted standalone deployment behind a reverse proxy, Next.js constructs the middleware request URL from the internal listener address. As a result, visiting `/` could redirect users to an internal URL such as `https://localhost:3000/login`, even when the proxy forwards the public host and protocol correctly.

Use the configured `NEXT_PUBLIC_BASE_URL` as the origin for the self-hosted login redirect. This is already Kan's canonical installation URL and is also used by the authentication setup and other absolute links. The request URL remains the fallback when the public base URL is missing, empty, or malformed at runtime.

This deliberately does not derive the redirect target from `Host` or `X-Forwarded-Host`, which would require an explicit trusted-proxy boundary and could otherwise allow host-header-driven redirects.

## Reproduction

Run the standalone server behind a TLS-terminating reverse proxy and request `/` with a public host:

`curl -I -H 'Host: kan.example.com' -H 'X-Forwarded-Proto: https' http://127.0.0.1:3000/`

Before this change, the response points to the internal listener:

`Location: https://localhost:3000/login`

After this change, with `NEXT_PUBLIC_BASE_URL=https://kan.example.com`:

`Location: https://kan.example.com/login`

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

## Testing

- `pnpm --filter @kan/web test` — 12 tests passed
- Prettier check for the changed files
- ESLint check for the changed files
- `git diff --check upstream/main...HEAD`
- reproduced against an unmodified standalone build behind forwarded host/protocol headers
- regression coverage for a runtime `NEXT_PUBLIC_BASE_URL` without a URL scheme

## Linked issue

Not applicable: this is a self-hosting bug fix and does not change intended product behavior.

### DIFF
--- a/apps/web/src/middleware.ts
+++ b/apps/web/src/middleware.ts
@@ -5,7 +5,11 @@ import { env } from "next-runtime-env";
 export function middleware(request: NextRequest) {
   if (request.nextUrl.pathname === "/") {
     if (env("NEXT_PUBLIC_KAN_ENV") !== "cloud") {
-      const loginUrl = new URL("/login", request.url);
+      const publicBaseUrl = env("NEXT_PUBLIC_BASE_URL");
+      const loginUrl = new URL(
+        "/login",
+        publicBaseUrl?.length ? publicBaseUrl : request.url,
+      );
       return NextResponse.redirect(loginUrl);
     }
   }
--- a/apps/web/src/middleware.ts
+++ b/apps/web/src/middleware.ts
@@ -2,15 +2,24 @@ import type { NextRequest } from "next/server";
 import { NextResponse } from "next/server";
 import { env } from "next-runtime-env";
 
+function resolveLoginUrl(request: NextRequest) {
+  const publicBaseUrl = env("NEXT_PUBLIC_BASE_URL");
+
+  if (publicBaseUrl?.length) {
+    try {
+      return new URL("/login", publicBaseUrl);
+    } catch {
+      // Runtime-injected values may bypass the build-time environment schema.
+    }
+  }
+
+  return new URL("/login", request.url);
+}
+
 export function middleware(request: NextRequest) {
   if (request.nextUrl.pathname === "/") {
     if (env("NEXT_PUBLIC_KAN_ENV") !== "cloud") {
-      const publicBaseUrl = env("NEXT_PUBLIC_BASE_URL");
-      const loginUrl = new URL(
-        "/login",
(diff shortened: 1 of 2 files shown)


apps/web/src/middleware.ts
```diff
@@ -5,7 +5,11 @@ import { env } from "next-runtime-env";
 export function middleware(request: NextRequest) {
   if (request.nextUrl.pathname === "/") {
     if (env("NEXT_PUBLIC_KAN_ENV") !== "cloud") {
-      const loginUrl = new URL("/login", request.url);
+      const publicBaseUrl = env("NEXT_PUBLIC_BASE_URL");
+      const loginUrl = new URL(
+        "/login",
+        publicBaseUrl?.length ? publicBaseUrl : request.url,
+      );
```
hjball:
```suggestion
if (publicBaseUrl?.length) {
      try {
        return new URL("/login", publicBaseUrl);
      } catch {
        // fall back to request origin
      }
    }

    return new URL("/login", request.url);
  }

  export function middleware(request: NextRequest) {
    if (request.nextUrl.pathname === "/") {
      if (env("NEXT_PUBLIC_KAN_ENV") !== "cloud") {
        return NextResponse.redirect(resolveLoginUrl(request));
      }
    }
```

We should include a fallback in case we get a malformed base URL (missing scheme for example). Something like this ^^

Otherwise lgtm @bionicman

hjball (changes requested):

apps/web/src/middleware.ts
bionicman: Addressed in 62ecfd09. I moved the URL selection into `resolveLoginUrl`, guarded the configured base URL, and added a regression test for a runtime value without a scheme. Although `NEXT_PUBLIC_BASE_URL` is declared as `z.string().url()` in `src/env.ts`, this middleware reads it through `next-runtime-env`, so a runtime-injected value can bypass the build-time schema. The fallback keeps a bad deployment value from turning `/` into a middleware exception. It still uses the request origin only when the setting is missing, empty, or malformed, and does not introduce trust in forwarded host headers. The full web suite passes (12 tests), as do targeted ESLint and Prettier checks. Thanks for catching this — ready for another look.

hjball (approved):
Nice one @bionicman

Issue #568, opened 2026-09-01
Move cards between boards


nickmeinhold:
## ✨ Feature Request

**Is your feature request related to a problem? Please describe.**

Asked for in the Discord #help thread "Moving cards between board": a general to-do board they want to split into various other boards, with no way to move the cards across.

**Describe the solution you'd like**

Let the card's existing "Move to another list" picker offer lists on the workspace's other boards, grouped by board.

As far as I can tell this is a UI change only, since the API already does both halves:

- card.update accepts any listPublicId in the workspace and does not check that the list is on the card's board, so it already performs a cross-board move
- board.all already returns every board in the workspace with its lists and its labels, which is exactly the data the picker needs

The scoping is two lines:

- CardContextMoveListModal.tsx:54, const lists = card?.list?.board?.lists ?? []
- CardContextDuplicateModal.tsx:56, const lists = board?.lists ?? []

**Describe alternatives you've considered**

- Leave it to the API and MCP. It works today, but it is not reachable for a non-technical user, and what is being described is ordinary board reorganisation.
- Copy instead of move. Simpler, but it leaves the original behind and is not what was asked for.
- A dedicated "move card to board" dialog separate from the list picker. More discoverable, but it duplicates a control that already exists.

**Additional context**

The part that needs a decision is labels rather than the picker. Labels are board-scoped (labels.boardId), so a card arriving on a new board brings labels belonging to its old one. Options, none obviously right:

1. Drop labels the destination board does not have
2. Match by name and link the destination equivalent, dropping the rest
3. Create the missing labels on the destination board
4. Block the move while the card holds labels the destination lacks

Since `board.all` already returns each board's labels, whichever rule you choose can be appthe page is alreadyfetching.
                                                       
Worth flagging that this is already reachable today rather than hypothetical: because card.update has no board check, a cross-board move can be made through the API right now, and the card keeps its old board's labels afterwards. So there is an existing gap here whether or not the picker changes. I have a fix for the writer side of it on a branch and am happy to open it separately, though the question above is yours to answer first.

**Implementation ideas (optional)**

Swap those two list sources for a board.all query and group the options by board name. Members and checklists are workspace-scoped or card-owned, so they travel without needing a rule.

**Would you like to work on this feature?**

- [x] Yes, I'd like to help implement this feature
- [ ] No, I'm just suggesting the feature

hjball:
Sounds good to me @nickmeinhold

PR #569, opened 2026-09-01
fix(cards): keep card labels within the card's own board

nickmeinhold:
## Description

Labels are board-scoped (`labels.boardId`), but `_card_labels` stores only `(cardId, labelId)`, so the schema cannot reject a card on one board holding a label from another. Every permission check on these paths is *workspace*-scoped, and a workspace holds many boards.

Five writers touch the join table (`insert(cardsToLabels)`). Four could produce an invalid link:

| Writer | Before |
|---|---|
| `card.create` | inserts any `labelPublicIds` unchecked |
| `card.duplicate` | target list need only share a workspace, so a cross-board copy carried the source board's labels |
| `card.addOrRemoveLabel` | resolves a label by publicId, links it with no board comparison |
| `labelRepo.create` (with a `cardId`) | creates a label on one board, links it to a card on another |

The importer is already correct, because it creates fresh labels on the board it builds.

Reachable today through the public API and MCP, not through the UI, because the pickers are board-scoped.

**The fix** is enforced in the repository writers rather than the routers, so it holds for future callers. The check and the insert share one transaction with the card rows locked (`FOR UPDATE OF card`, ordered by id), because a card's board is a property of whatever list its `listId` points at.

Also fixed along the way:

- `card.create` validated labels *after* creating the card, so a bad label meant an orphaned card and a 500. It validates first and returns `BAD_REQUEST`.
- `card.duplicate` had the same ordering problem, and retrying as its own error message suggested produced a second card.
- A partial label lookup counted as success, silently creating a card with fewer labels than requested.
- `CardLabelBoardError` carries a closed violation type, so "unknown card", "unknown label" and "board mismatch" stop sharing one string.

**Known limits, stated rather than implied:**

- `card.update` can still move a card to another board, leaving the labels it already holds invalid. No join row is written, so a writer guard cannot see it. What should happen to those labels (drop, match by name, recreate, block the move) is a product decision and the same question as #568, so I have not answered it here.
- The `FOR UPDATE` lock is not covered by a test. Removing it leaves all 12 green. Proving it needs a contention test.

Happy to take this in a different direction if you would rather the invariant live in the schema.

## Reviewing

Most of the diff is the new test file. `label.repo.ts` reads as ~60 lines but is 15 insertions and 2 deletions ignoring whitespace, the rest being re-indentation from wrapping the body in a transaction.

https://github.com/kanbn/kan/pull/569/files?w=1

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

- [ ] I have included screenshots for any UI changes — no UI changes in this PR

## Linked issue

Relates to #568 (deliberately not `Closes`: that issue asks for cross-board card moves, which this PR does not implement. It fixes the label scoping that feature would otherwise expose.)

---

12 integration tests against real Postgres. Every cross-board assertion was confirmed red before its fix, with same-board positive controls green throughout.


nickmeinhold:
The failing `test` check here is #593, not something in this PR. `board-move.test.ts` pays the cost of importing better-auth and Drizzle in its first test and lands near the 5000ms timeout on CI.

PR #570, opened 2026-09-02
perf: deduplicate avatar URL generation in API responses

bionicman:
## Description

`board.byId` generated a presigned avatar URL for every workspace member and
every card-member occurrence. The same image key can appear on hundreds or
thousands of cards, so one board request repeatedly created S3 clients and
signed identical URLs.

This change adds a procedure-scoped resolver that caches the promise for each
image key. Concurrent duplicates share the same work. Resolver instances are
not shared between procedures or requests, so URL expiry and access behaviour
remain unchanged.

Following review feedback, the same pattern is now used by `workspace.byId`,
`card.byId`, and `card.getActivities`. The activity feed shares one resolver
between activity actors and member users. The two user endpoints are left as
direct calls because each resolves only one avatar and cannot contain duplicate
keys.

## Production reference

On a migrated board with 1,890 cards and 4,600 card-member assignments, 4,196
avatar-bearing occurrences referenced only 33 unique image keys. Populating
those avatars changed `board.byId` from roughly sub-second responses to a 4–12
second range and delayed unrelated procedures in the same tRPC batch. With this
change, that procedure performs at most one signing operation per unique key.

The review follow-up was also A/B tested using two isolated builds against the
same production data:

- an activity feed with 46 avatar occurrences and 3 unique keys reduced its
  median server duration from 90.5 ms to 25 ms;
- a second feed with 31 raw avatar occurrences, including member-user avatars,
  and 3 unique keys reduced its median from 73.5 ms to 25 ms;
- the tested workspace and card detail contained 108 avatar occurrences and
  105 unique keys, so they showed no material timing change, as expected.

Normalized responses were identical and payload sizes did not change.

## Tests

The tests cover:

- concurrent reuse of duplicate image keys;
- isolation between resolver instances;
- duplicate workspace-member avatars in workspace and card responses;
- reuse across activity actors and member users.

Tested with:

- `pnpm --filter @kan/api test -- src/routers/card-members.test.ts src/routers/avatar-urls.test.ts src/utils/avatarUrls.test.ts`
- ESLint and Prettier checks for the changed files

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

## Linked issue

### DIFF
--- a/packages/api/src/routers/avatar-urls.test.ts
+++ b/packages/api/src/routers/avatar-urls.test.ts
@@ -1,0 +1,231 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import * as cardRepo from "@kan/db/repository/card.repo";
+import * as cardActivityRepo from "@kan/db/repository/cardActivity.repo";
+import * as workspaceRepo from "@kan/db/repository/workspace.repo";
+import { generateAvatarUrl } from "@kan/shared/utils";
+
+import { assertPermission } from "../utils/permissions";
+
+vi.mock("@kan/db/repository/card.repo", () => ({
+  getWorkspaceAndCardIdByCardPublicId: vi.fn(),
+  getWithListAndMembersByPublicId: vi.fn(),
+}));
+vi.mock("@kan/db/repository/cardActivity.repo", () => ({
+  getPaginatedActivities: vi.fn(),
+}));
+vi.mock("@kan/db/repository/cardComment.repo", () => ({}));
+vi.mock("@kan/db/repository/checklist.repo", () => ({}));
+vi.mock("@kan/db/repository/label.repo", () => ({}));
+vi.mock("@kan/db/repository/list.repo", () => ({}));
+vi.mock("@kan/db/repository/subscription.repo", () => ({}));
+vi.mock("@kan/db/repository/workspace.repo", () => ({
+  getByPublicIdWithMembers: vi.fn(),
+}));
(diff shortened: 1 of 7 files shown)


hjball (changes requested):
Great stuff @bionicman 

Not a high-priority but please can we adopt this pattern in `workspace.ts` and `card.ts` too

bionicman:
> Great stuff @bionicman
> 
> Not a high-priority but please can we adopt this pattern in `workspace.ts` and `card.ts` too

Deal

bionicman:
@hjball Done — the resolver is now used by `workspace.byId`, `card.byId`, and `card.getActivities`, with one shared instance for activity actors and member users. I rebased the branch onto the latest `main` and added router-level regression coverage for all three procedures.

The production-shaped A/B check showed the material benefit in activity feeds: 46 avatar occurrences / 3 unique keys went from a 90.5 ms to 25 ms median server duration, and a second fixture covering member-user avatars went from 73.5 ms to 25 ms. Workspace and card detail had 108 occurrences / 105 unique keys, so their timing was effectively unchanged, as expected. Normalized responses and payload sizes were identical.

One small rebase compatibility fix is included: after #582, the existing card router test mock was missing the new `normalizeDescription` export, which the new unit-test workflow from #583 now exercises. Ready for another look.

hjball (approved):
Nice one, cheers @bionicman

PR #572, opened 2026-09-03
perf: reduce board payload with compact card summaries

bionicman:
## Title

perf: reduce board payload with compact card summaries

## Description

The authenticated board view only renders a small set of card indicators, but
`board.byId` currently loads and serializes full descriptions, comments,
attachments, checklists, and checklist items for every card. This becomes
expensive on large imported boards even after the relation lookups themselves
are indexed.

This adds an opt-in `cardView: "summary"` read model and uses it for the board
UI. Each card receives only the aggregate values needed for its description,
comment, attachment, and checklist indicators. The existing `"full"` mode
remains the default, preserving the response contract for existing API
consumers and the public board view.

The change also removes redundant board requests from the new-card and
duplicate-card flows. A proposed 30-second board query `staleTime` was removed
after E2E testing showed that it could hide labels created from the card view.

The issue was reproduced on imported self-hosted data containing 38 boards and
14,524 cards. Representative active boards contained 1,203, 1,587, 1,890, and
2,511 cards.

For two large boards, the compressed `board.byId` response dropped from
approximately 510–954 KB to 98–216 KB. End-to-end requests through the reverse
proxy dropped from roughly 2.15–3.24 seconds to 1.29–1.60 seconds.

`EXPLAIN (ANALYZE, BUFFERS)` on the 1,890-card board confirmed index scans for
card attachments, comments, checklists, and checklist items. Consolidating the
two checklist counters into one aggregate reduced the representative summary
query from 75.6 to 67.4 ms and from 19,872 to 15,521 shared buffer hits.

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

## Testing

- `git diff --check upstream/main...HEAD`
- `pnpm --filter @kan/web test` — 8 tests passed
- `pnpm --filter @kan/api test -- board-summary.integration.test.ts`
- production build completed successfully
- focused self-hosted Playwright coverage passed for compact card details,
  board filters, card move/delete, and card duplication
- PostgreSQL execution plans verified on a 1,890-card board

No screenshots are included because the change does not intentionally alter
the rendered UI.

## Linked issue

Not applicable: this is a board loading performance bug fix and does not change
product behavior.

### DIFF
--- a/packages/db/src/repository/board.repo.ts
+++ b/packages/db/src/repository/board.repo.ts
@@ -272,7 +272,7 @@ export const getByPublicId = async (
               ).as("description"),
               hasDescription: (includeCardDetails
                 ? sql<boolean>`FALSE`
-                : sql<boolean>`COALESCE(BTRIM(REGEXP_REPLACE(COALESCE(${card.description}, ''), '<[^>]*>', '', 'g')), '') <> ''`
+                : sql<boolean>`${card.description} IS NOT NULL`
               ).as("has_description"),
               attachmentCount: (includeCardDetails
                 ? sql<number>`0`
--- a/packages/db/src/repository/board.repo.ts
+++ b/packages/db/src/repository/board.repo.ts
@@ -10,6 +10,7 @@ import {
   isNull,
   lt,
   or,
+  sql,
 } from "drizzle-orm";
 
 import type { dbClient } from "@kan/db/client";
@@ -156,8 +157,10 @@ export const getByPublicId = async (
     lists: string[];
     dueDate: DueDateFilter[];
     type: "regular" | "template" | undefined;
+    cardView?: "full" | "summary";
   },
 ) => {
+  const includeCardDetails = filters.cardView !== "summary";
   let cardIds: string[] = [];
 
   if (filters.labels.length > 0 || filters.members.length > 0) {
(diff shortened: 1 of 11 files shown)


bionicman:
The failing unit-test check is caused by an interaction already present on current `main`: #576 fully mocks `@kan/shared/utils`, while #582 added a call to `normalizeDescription` in the tested router. The focused fix is isolated in #585; this PR's build and E2E checks pass.

Once #585 lands, I’ll rebase this branch so the unit-test check reruns against the fixed baseline.

packages/db/src/repository/board.repo.ts
```diff
@@ -254,12 +257,42 @@ export const getByPublicId = async (
...
               publicId: true,
               title: true,
-              description: true,
+              description: false,
               listId: true,
               index: true,
               dueDate: true,
               cardNumber: true,
             },
+            extras: (card) => ({
```
hjball, line 266: This is my only concern with this PR - I think we can simplify this.

I've just [merged a PR](https://github.com/kanbn/kan/pull/582) to normalise descriptions which will prevent empty html/white space from being set on descriptions going forward. Once deployed I'll clean up existing data so we can safely check `description IS NULL`

hjball (changes requested):
One comment that needs addressing otherwise lgtm

bionicman:
Good point — I’ve simplified this to rely on `description IS NOT NULL` and pushed the update.

While tracing the invariant introduced in #582, I noticed that `card.update` still passed editor HTML directly to the repository. Clearing an existing description could therefore store empty HTML such as `<p></p>` instead of `NULL`, so I opened #592 to address that omission separately.

The two PRs can be reviewed in parallel, but I think it is cleaner to merge #592 first so the write-side invariant is in place before #572 starts relying on it.

packages/db/src/repository/board.repo.ts
bionicman: Addressed in `7e7a368`: the query now relies on `description IS NOT NULL`. #592 has since merged, so card updates follow the same normalised-description invariant. All checks are passing.

bionicman:
Rebased onto current `main`; the requested simplification is retained and all checks are green. Ready for re-review.

bionicman:
@hjball The requested change has been addressed, the branch is rebased, and CI is green. Could you take another look when you have a chance?

PR #574, opened 2026-09-03
fix: deduplicate mention emails per content change

bionicman:
## Description

Mention email deduplication currently uses `userId + type + cardId`. Once a
notification row exists, the same member will not receive another mention email
from that card, even when they are explicitly mentioned in a new comment.

The notification row is also created before SMTP delivery. If that delivery
fails, the row remains and suppresses later attempts even though no email was
sent.

This changes the boundary to newly added mentions in each content change:

- new descriptions and comments notify each mentioned member once;
- edits notify only members added relative to the previous HTML;
- editing text around an existing mention does not send another email;
- a mention in another comment, or one removed and later added again, is a new
  event;
- self mentions and pending members remain excluded.

SMTP delivery still runs fire-and-forget after the content mutation. A
notification row is recorded only after SMTP accepts the message. This keeps the
existing lightweight delivery model: there is no queue or automatic retry, and
a process failure between SMTP acceptance and recording can rarely result in a
duplicate.

The original mention email behavior was introduced in #372.

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

- [ ] I have linked the related issue below — no separate issue; related to #372

## Testing

- 10 focused API tests cover old/new mention sets, separate content events,
  self/pending members, SMTP failures and SMTP-before-notification ordering.
- A self-hosted Playwright scenario exercises the editor, tRPC, PostgreSQL and
  Mailpit. It covers the initial mention, an edit that preserves the mention,
  and removal followed by re-adding it.
- API and E2E typechecks and focused lint pass.

## Linked issue

No separate issue. Related implementation: #372.

### DIFF
--- a/packages/api/src/routers/card.ts
+++ b/packages/api/src/routers/card.ts
@@ -173,13 +173,12 @@ export const cardRouter = createTRPCRouter({
       }
 
       if (input.description) {
-        sendMentionEmails({
+        void sendMentionEmails({
           db: ctx.db,
           cardPublicId: newCard.publicId,
-          commentHtml: input.description,
+          previousHtml: null,
+          nextHtml: input.description,
           commenterUserId: userId,
-        }).catch((error) => {
-          console.error("Failed to send mention emails:", error);
         });
       }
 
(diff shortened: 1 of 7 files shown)


hjball (approved):
nice work @bionicman

Issue #416, opened 2026-02-24
Security: Cross-workspace resource association via missing workspace validation on sub-resource lookups


lighthousekeeper1212:
## Summary

Multiple tRPC card procedures verify workspace-level permission against the primary resource (card) but fail to verify that secondary resources (lists, labels, members, comments) belong to the same workspace. This is a defense-in-depth issue - the nanoid publicIds make exploitation impractical, but proper workspace boundary validation should be added.

### 1. card.update - Target list not workspace-validated
`packages/api/src/routers/card.ts`: When updating with `listPublicId`, the target list is looked up via `listRepo.getByPublicId()` which queries `WHERE publicId = ?` without workspace filter. No check that the target list belongs to the same workspace as the card.

### 2. card.addOrRemoveLabel - Label not workspace-validated
Same file: `labelRepo.getByPublicId()` queries by publicId only. No check the label belongs to the same board/workspace as the card.

### 3. card.addOrRemoveMember - Member not workspace-validated
Same file: `workspaceRepo.getMemberByPublicId()` queries by publicId only. No check the member belongs to the same workspace.

### 4. card.updateComment / card.deleteComment - Comment not card-validated
Comment looked up by `commentPublicId` without verifying it belongs to the specified `cardPublicId`. Permission check uses the provided card's workspace, but the comment could belong to a different card/workspace.

### 5. card.create - Same issue with label/member arrays
Labels and members resolved by publicId arrays without workspace validation.

## Mitigating Factor

PublicIds are 12-character nanoid with 36-char alphabet (~62 bits of entropy). Brute-force is impractical. However, users who are members of multiple workspaces can obtain publicIds legitimately.

## Recommended Fix

After looking up secondary resources by publicId, verify they belong to the same workspace/board as the primary resource. For comments, verify `existingComment.cardId === card.id`.

CWE-639: Authorization Bypass Through User-Controlled Key (LOW severity)

bionicman:
We ran into the member-assignment part of this while reviewing how paused
workspace members behave on cards. It is still present on current `main`:

- `card.create` resolves `memberPublicIds` without scoping them to the card's
  workspace;
- `card.addOrRemoveMember` resolves the member by public ID and checks the
  caller's permission against the card workspace, but does not verify that the
  selected member belongs to it.

This is the same boundary problem fixed for member and permission endpoints in
#410 / #413, but those changes did not cover card assignment.

@hjball, we can submit a small PR for the two member-assignment paths with
[...]

Issue #410, opened 2026-02-22
Security: Cross-Workspace Member IDOR in Member and Permission Routers (CWE-639)


lighthousekeeper1212:
## Summary

Multiple endpoints in the member and permission tRPC routers are vulnerable to cross-workspace IDOR. An authenticated user who is an admin of Workspace A can operate on members of Workspace B, including deleting members, changing roles, and granting/revoking permissions.

**Severity:** HIGH (cross-workspace member management, role manipulation, permission escalation)

## Root Cause

`memberRepo.getByPublicId()` in `packages/db/src/repository/member.repo.ts` (line 54) queries the `workspace_members` table by `publicId` only, without filtering by `workspaceId`:

```typescript
export const getByPublicId = async (db: dbClient, publicId: string) => {
  return db.query.workspaceMembers.findFirst({
    where: eq(workspaceMembers.publicId, publicId),
    // Missing: workspaceId filter
  });
};
```

The authorization flow in all affected endpoints:
1. Resolve workspace from `input.workspacePublicId` (attacker's own workspace)
2. Call `assertPermission(db, userId, workspace.id, ...)` — passes because attacker is admin of their own workspace
3. Fetch target member with `memberRepo.getByPublicId(db, input.memberPublicId)` — fetches ANY member from ANY workspace
4. Operate on the member without checking `member.workspaceId === workspace.id`

## Affected Endpoints (6)

| Endpoint | File | Impact |
|----------|------|--------|
| `member.delete` | `packages/api/src/routers/member.ts:172` | Cross-workspace member deletion |
| `member.updateRole` | `packages/api/src/routers/member.ts:674` | Cross-workspace role manipulation |
| `permission.getMemberPermissions` | `packages/api/src/routers/permission.ts:76` | Cross-workspace permission info disclosure |
| `permission.grantPermission` | `packages/api/src/routers/permission.ts:164` | Cross-workspace permission escalation |
| `permission.revokePermission` | `packages/api/src/routers/permission.ts:232` | Cross-workspace permission revocation |
| `permission.resetMemberPermissions` | `packages/api/src/routers/permission.ts:297` | Cross-workspace permission override reset |

## Secure Comparison

Resource endpoints (boards, cards, lists, labels, checklists, attachments) all derive `workspaceId` FROM the target resource via FK chain, ensuring correct scoping:

```typescript
// board.ts - workspace derived from board itself
[...]

PR #576, opened 2026-09-04
fix: scope card member assignments to the workspace

bionicman:
## Description

Card member IDs were resolved globally in the create and add/remove paths. A
caller with permission to edit a card could therefore supply a membership ID
from another workspace.

This change derives the workspace from the target card or list and uses it for
all member lookups. Card creation validates the complete member set before
writing the card, so a mixed valid/foreign request fails atomically. Card
duplication applies the same boundary as defense in depth for legacy or
otherwise inconsistent relationships. Soft-deleted memberships are excluded
from these assignment lookups as well.

The behavior is covered for card creation, add/remove and duplication.

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

- [x] I have included screenshots for any UI changes (not applicable)

## Testing

- `pnpm --filter @kan/api test -- card-members.test.ts`
- `pnpm --filter @kan/api exec tsc --noEmit --jsx preserve`

## Linked issue

Refs #416. This intentionally addresses only the card-member assignment part
of that report, so it does not close the broader issue.

The same workspace-boundary class was previously addressed for member and
permission endpoints in #410 / #413.

### DIFF
--- a/packages/api/src/routers/card-members.test.ts
+++ b/packages/api/src/routers/card-members.test.ts
@@ -1,0 +1,251 @@
+import type { TRPCError } from "@trpc/server";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import * as cardRepo from "@kan/db/repository/card.repo";
+import * as cardActivityRepo from "@kan/db/repository/cardActivity.repo";
+import * as listRepo from "@kan/db/repository/list.repo";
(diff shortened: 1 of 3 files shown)


hjball (approved):
good stuff @bionicman

PR #577, opened 2026-09-04
fix: preserve custom label colours when editing

bionicman:
## Description

Labels created through the API or MCP can use a colour outside the built-in
eight-colour palette. When one of those labels was opened for editing,
`LabelForm` could not resolve a selected option and attempted to render an
undefined value.

This keeps the existing creation palette unchanged while treating the stored
HEX value as the current option when it is not one of the presets. The label can
then be opened and saved without losing its colour, or changed to one of the
normal presets.

The shared resolver is covered for preset, custom and missing colour values.
The existing self-hosted label scenario now also assigns a custom colour via
the API, opens the edit form and saves it again.

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

There is no intended layout or design change; this restores the edit form for
an already supported stored value.

## Testing

- `pnpm --filter @kan/web test` — 11 tests passed
- targeted ESLint and Prettier checks for the changed files
- self-hosted `card-labels.spec.ts` against PostgreSQL 16 — 1 test passed
- Next.js production build completed as part of the Playwright run

## Linked issue

Not applicable: this fixes editing for custom label colours already accepted by
the API, database and MCP.

### DIFF
--- a/apps/web/src/components/LabelForm.tsx
+++ b/apps/web/src/components/LabelForm.tsx
@@ -4,6 +4,7 @@ import { Fragment, useEffect } from "react";
 import { Controller, useForm } from "react-hook-form";
 import { HiChevronUpDown, HiXMark } from "react-icons/hi2";
 
+import type { Colour } from "@kan/shared/constants";
 import { colours } from "@kan/shared/constants";
 
 import Button from "~/components/Button";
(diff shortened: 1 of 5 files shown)


bionicman:
A little context for why this general fix came up: I have a follow-up ready for the Trello importer. It preserves Trello's label colours instead of assigning Kan colours by array position, and explicitly requests  because Trello otherwise limits nested labels to 50.

That importer can create HEX values outside Kan's built-in editing palette, so this PR is the small general prerequisite: imported labels should remain editable without losing their stored colour.

I'll open and link the importer change as a separate PR after this lands, so the scope here stays focused.

hjball (approved):
lgtm @bionicman

PR #579, opened 2026-09-04
fix: preserve Trello label colours during import

bionicman:
## Summary

- preserve Trello label colours instead of assigning colours by array position
- support all 30 named Trello colour variants, with explicit fallbacks for
  colourless and unknown values
- request up to 1,000 labels per board instead of relying on Trello's default
  nested collection limit
- cover the mapping and import request in unit and self-hosted E2E tests

## Why

We are migrating an established Trello workspace rather than a small example
board. It currently has 422 label definitions; the two largest boards have 75
and 57. There are 77 dark/light colour definitions used by 336 card-label
assignments. The current importer both truncates large label collections and
loses those source colours.

The colour mapping stays inside the Trello import boundary. Kan's shared label
palette and creation UI are unchanged. Colourless labels retain the existing
grey fallback, and unknown future Trello values use a deterministic fallback.

This follows #577, which made imported custom HEX values safe to edit.

## Testing

- `pnpm --filter @kan/api exec vitest run src/utils/trello.test.ts` — 34 passed
- API and E2E typechecks
- package-scoped ESLint and Prettier checks
- both self-hosted Trello import scenarios, including a `red_dark` label and a
  mock assertion for `labels_limit=1000`

### DIFF
--- a/packages/api/src/routers/import.ts
+++ b/packages/api/src/routers/import.ts
@@ -11,7 +11,6 @@ import * as labelRepo from "@kan/db/repository/label.repo";
 import * as listRepo from "@kan/db/repository/list.repo";
 import * as workspaceRepo from "@kan/db/repository/workspace.repo";
 import { createLogger } from "@kan/logger";
-import { colours } from "@kan/shared/constants";
 import {
   generateSlug,
   generateUID,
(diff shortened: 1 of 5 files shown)


hjball (commented):
We don't currently support the full Trello colour range so it won't be possible to assign the imported colours via the Kan UI.

I don't think that's an issue but worth considering before we merge.

Otherwise lgtm

bionicman:
The failing unit-test check is caused by an interaction already present on current `main`: #576 fully mocks `@kan/shared/utils`, while #582 added a call to `normalizeDescription` in the tested router. The focused fix is isolated in #585; this PR's build and E2E checks pass.

Once #585 lands, I’ll rebase this branch so the unit-test check reruns against the fixed baseline.

hjball (approved):
lgtm @bionicman

PR #581, opened 2026-09-04
feat: clarify active board filters

bionicman:
## Description

Board filters currently show only a total count on the Filter button. The
dropdown does not indicate which groups contain selected values, and clearing
filters is hidden behind a hover interaction on the count badge.

This change:

- shows the selected-value count next to each filter group;
- keeps the total badge as a status indicator instead of a hidden action;
- adds an explicit `Clear filters` action at the bottom of the menu;
- adds a back row for moving between filter groups without reopening the menu;
- extends the existing board-filter E2E scenario to cover the count, selected
  checkbox and clear action.

Filter state and behavior are unchanged. The existing URL query remains the
source of truth, and the shared dropdown only gains optional presentation
slots.

## Type of change

- [ ] Bug fix
- [x] Feature (requires an approved issue — see below)
- [ ] Refactor / chore
- [ ] Documentation

## Screenshots

Group counts and explicit reset, desktop:

Filter group counts on desktop

Selected value inside a group, desktop:

Selected filter value on desktop

Group counts and explicit reset, mobile:

Filter group counts on mobile

## Checklist

- [x] I have linked the related issue below
- [x] My code follows the existing style and conventions
- [x] I have tested my changes locally
- [x] I have included screenshots for any UI changes

## Linked issue

Closes #580

PR #585, opened 2026-09-04
test: update card member mock for description normalization

bionicman:
## What changed

Adds `normalizeDescription` to the shared-utils mock used by the card-member router tests.

## Why

The card-member tests introduced in #576 fully mock `@kan/shared/utils`. After #582 began normalising descriptions in card create and duplicate, those tests fail before reaching their workspace-scoping assertions because the mock does not expose the new helper.

This is test-only and does not change runtime behaviour.

It also restores the unit-test baseline currently blocking otherwise passing
#572, #579, #581, and #586.

## Verification

`pnpm --filter @kan/api exec vitest run src/routers/card-members.test.ts` — 5 tests passed.


hjball (approved):
Ah nice one, I'll close mine

hjball (approved):

PR #587, opened 2026-09-04
fix(db): enable pg_trgm in PGlite

bionicman:
The PGlite fallback only registers uuid-ossp, while the fuzzy search migration creates the pg_trgm extension. Starting Kan without POSTGRES_URL can therefore reject the migration because PGlite does not know that extension. In the API test suite this can surface later as an unhandled rejection.

This change registers pg_trgm in the runtime fallback and shares the same extension registry with the integration-test database so the two configurations cannot silently drift again.

Tested with:

- pnpm --filter @kan/db typecheck
- pnpm --filter @kan/api exec vitest run integration-tests/webhook.integration.test.ts (14 tests)


hjball (approved):
Good spot @bionicman

PR #590, opened 2026-09-05
fix: validate attachment download URLs when S3_ENDPOINT is unset

nickmeinhold:
## Description

The URL allowlist added in #432 for CVE-2026-32255 runs inside `if (s3Endpoint)`. S3_ENDPOINT is optional, and .env.example recommends leaving it blank for AWS S3, so on that configuration no validation ran and the endpoint still fetched any user-supplied URL unauthenticated.

Attachment URLs are presigned S3 URLs from `generateAttachmentUrl`, so the allowlist is now built from S3_ENDPOINT and NEXT_PUBLIC_STORAGE_URL, whichever are set, and the request is refused when neither is. Subdomain matching is kept because virtual-hosted style puts the bucket in a subdomain of the endpoint host.

## Type of change

Behaviour change: deployments with neither variable set will now get a 403 from this endpoint instead of an unvalidated fetch.

### DIFF
--- a/apps/web/src/utils/attachmentDownload.ts
+++ b/apps/web/src/utils/attachmentDownload.ts
@@ -1,0 +1,45 @@
+/**
+ * Attachment URLs are presigned S3 URLs produced by generateAttachmentUrl, so
+ * their host is whichever endpoint the S3 client resolved. Virtual-hosted style
+ * puts the bucket in a subdomain, so subdomains of an allowed host are accepted.
+ *
+ * Returns null when a configured value is not a valid URL.
+ */
+export const getAllowedAttachmentHosts = (
+  s3Endpoint: string | undefined,
+  storageUrl: string | undefined,
+): string[] | null => {
+  const hosts: string[] = [];
+
+  for (const candidate of [s3Endpoint, storageUrl]) {
+    if (!candidate) continue;
+    try {
+      hosts.push(new URL(candidate).hostname.toLowerCase());
+    } catch {
+      return null;
+    }
+  }
+
+  return hosts;
+};
+
+export const isAttachmentUrlAllowed = (
+  url: string,
+  allowedHosts: string[],
+): boolean => {
+  if (!allowedHosts.length) return false;
+
+  let parsed: URL;
+  try {
+    parsed = new URL(url);
+  } catch {
+    return false;
+  }
+
+  const hostname = parsed.hostname.toLowerCase();
+
+  return allowedHosts.some(
+    (allowedHost) =>
+      hostname === allowedHost || hostname.endsWith(`.${allowedHost}`),
+  );
+};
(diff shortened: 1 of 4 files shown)


apps/web/src/utils/attachmentDownload.ts
```diff
@@ -0,0 +1,45 @@
...
+export const getAllowedAttachmentHosts = (
+  s3Endpoint: string | undefined,
+  storageUrl: string | undefined,
+): string[] | null => {
+  const hosts: string[] = [];
+
+  for (const candidate of [s3Endpoint, storageUrl]) {
+    if (!candidate) continue;
+    try {
+      hosts.push(new URL(candidate).hostname.toLowerCase());
```
hjball: I think something like `minio.internal:9000` will evaluate to an empty string here - we should guard against schemeless S3_ENDPOINTS

apps/web/src/utils/attachmentDownload.ts
```diff
@@ -0,0 +1,45 @@
...
+  let parsed: URL;
+  try {
+    parsed = new URL(url);
+  } catch {
+    return false;
+  }
+
+  const hostname = parsed.hostname.toLowerCase();
+
+  return allowedHosts.some(
```
hjball: I don't think this fully closes the gap - just matching on the subdomain will mean this endpoint can fetch any bucket on a shared domain such as`s3.us-east-1.amazonaws.com`

hjball (changes requested):
Couple of points/questions - let me know your thoughts @nickmeinhold

apps/web/src/utils/attachmentDownload.ts
nickmeinhold: Good catch, and it turned out worse than an empty string. `[""]` is non-empty so it skipped my "neither configured" guard, and then `endsWith("." + "")` matched any trailing-dot hostname, so `metadata.google.internal.` and `evil.com.` both got through.

Worth flagging that this one isn't only my patch. Main behaves the same way today, since `new URL("minio.internal:9000")` doesn't throw, `allowedHost` ends up `""`, and the existing check then passes anything ending in a dot. So it's a live path on any deploy with a schemeless endpoint, not just a gap in the PR.

Endpoints now need an explicit http(s) scheme, and an unusable one is refused rather than producing an empty host.

apps/web/src/utils/attachmentDownload.ts
nickmeinhold: Agreed, and the same holds for main right now, so any bucket on the configured domain is reachable today rather than this being only a gap in the patch.

I'd talked myself into keeping the wildcard because dropping it breaks virtual-hosted AWS, but the bucket name is what makes it safe to drop. It now requires our own bucket, either as the only subdomain label or as the first path segment, so sibling buckets are out. Blocked `evil.our-bucket.host` while I was in there too.

PR #591, opened 2026-09-05
fix(trello): protect stored access tokens

bionicman:
## Description

While preparing the stock Trello importer for an on-premise migration, we
noticed that Trello access tokens do not follow the storage pattern already
used by the GitHub integration. `saveGitHubToken` encrypts a token with Kan's
shared AES-GCM helper, and the GitHub importer decrypts it only when making an
API request. The Trello authentication endpoint currently stores the raw token,
and the Trello importer reads that raw value back.

This PR applies the existing GitHub approach to Trello. Newly connected Trello
tokens are encrypted with the same helper and `BETTER_AUTH_SECRET`, then
decrypted for Trello API requests.

Existing Trello installations need one additional compatibility step because
their stored tokens may still be plaintext. Encrypted Trello values therefore
have a versioned prefix: unprefixed values remain readable and are rewritten in
encrypted form the next time the integration is used. The rewrite has a
compare-and-swap condition so it cannot overwrite a token entered by a
concurrent reconnect. No database migration is required.

The integration providers endpoint also stops returning access and refresh
tokens. It retains the existing non-secret provider metadata. The GitHub token
save and use paths are otherwise unchanged.

We needed this behaviour in our fork before allowing users to connect Trello.
Since it is an extension of an existing Kan security pattern and is independent
of our migration tooling and product changes, we are contributing the same fix
upstream rather than keeping a private patch.

Tested with:

- `pnpm --filter @kan/api test -- trello-token.test.ts`
- `pnpm --filter @kan/api test -- integration-tests/integration.repo.integration.test.ts`
- the self-hosted Trello import Playwright tests
- the repository's unit, E2E and Docker build checks

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

## Linked issue

### DIFF
--- a/apps/web/src/pages/api/trello/authenticate.ts
+++ b/apps/web/src/pages/api/trello/authenticate.ts
@@ -4,7 +4,10 @@ import { addYears } from "date-fns";
 import { createNextApiContext } from "@kan/api/trpc";
 import { withApiLogging } from "@kan/api/utils/apiLogging";
 import { withRateLimit } from "@kan/api/utils/rateLimit";
-import { integrations } from "@kan/db/schema";
+import { encryptTrelloToken } from "@kan/api/utils/trello-token";
+import * as integrationsRepo from "@kan/db/repository/integration.repo";
+
+import { env } from "~/env";
 
 export default withRateLimit(
   { points: 100, duration: 60 },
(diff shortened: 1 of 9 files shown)


hjball (approved):
Changes look sensible to me @bionicman

PR #597, opened 2026-09-07
fix(board): prevent actions on optimistic lists

bionicman:
## Description

Creating a list adds it to the board cache immediately, before `list.create`
returns. Until that request completes, the list has a client-generated public
ID that does not exist on the server. On a slow connection, the UI currently
treats it like a persisted list, so adding or moving a card, renaming the list,
or moving the list itself can send that temporary ID to the API and fail with a
not-found error.

Kan already prefixes client-only IDs with `PLACEHOLDER_` for optimistic cards,
checklists, and checklist items. The board UI also uses that prefix to prevent
actions on cards that have not been acknowledged by the server yet. This change
extends the same established convention to optimistic lists rather than
introducing a separate pending-state mechanism.

Actions that require a persisted list stay disabled until the server
acknowledges the creation. The temporary list cannot be renamed, dragged,
deleted, used to create a card, or used as a card drop target. Once
`list.create` returns, the optimistic ID is replaced with the server ID and the
controls become available normally.

The existing rollback and query invalidation behaviour is unchanged. This does
not change the API or database schema.

A Playwright test delays `list.create`, verifies that actions stay unavailable
while the list is optimistic, releases the request, and then creates a card in
the acknowledged list.

Tested with:

- focused helper tests — 9 tests passed
- `@kan/e2e` typecheck
- ESLint and Prettier checks for the new E2E test
- focused self-hosted Playwright test — 1 test passed
- `git diff --check`

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

- [ ] I have included screenshots for any UI changes — no persistent visual change

## Linked issue

### DIFF
--- a/apps/web/src/utils/helpers.test.ts
+++ b/apps/web/src/utils/helpers.test.ts
@@ -5,9 +5,16 @@ vi.mock("next-runtime-env", () => ({
 }));
 
 import { env } from "next-runtime-env";
-import { getAvatarUrl } from "./helpers";
+import { getAvatarUrl, isPlaceholderPublicId } from "./helpers";
 
 const mockEnv = env as ReturnType<typeof vi.fn>;
+
+describe("isPlaceholderPublicId", () => {
+  it("identifies optimistic entity IDs", () => {
+    expect(isPlaceholderPublicId("PLACEHOLDER_abc123")).toBe(true);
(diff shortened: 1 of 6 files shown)


hjball (approved):
Nice one @bionicman

PR #610, opened 2026-09-09
fix(api): return valid tRPC rate limit errors

bionicman:
## Description

Rate-limited tRPC requests currently return a plain JSON message, which the tRPC client cannot parse as an error response. Return a standard `TOO_MANY_REQUESTS` envelope for both single and batch requests while keeping the generic rate-limit middleware reusable.

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

- [x] I have included screenshots for any UI changes (not applicable: no UI changes)

## Linked issue

N/A — bug fix without an existing issue.

## Verification

- Focused tests: 3 passed
- API typecheck
- Single and batch HTTP requests against a locally running app

### DIFF
--- a/apps/web/src/pages/api/trpc/[trpc].ts
+++ b/apps/web/src/pages/api/trpc/[trpc].ts
@@ -4,6 +4,7 @@ import { createNextApiHandler } from "@trpc/server/adapters/next";
 import { appRouter } from "@kan/api/root";
 import { createTRPCContext } from "@kan/api/trpc-context";
 import { withRateLimit } from "@kan/api/utils/rateLimit";
+import { createTRPCRateLimitResponder } from "@kan/api/utils/trpcRateLimit";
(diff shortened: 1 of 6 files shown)


hjball:
I think this has been fixed already https://github.com/kanbn/kan/pull/615, but let me know if otherwise and I can take another look @bionicman

bionicman:
Thanks — I checked this against #615. This PR is already based on the #615 merge commit, so both changes are present together. #615 fixes request identification for rate limiting and the MCP/API auth path, while the web tRPC endpoint still returns a generic `{ message }` body for HTTP 429. This PR returns the standard serialized tRPC `TOO_MANY_REQUESTS` error shape, including batched calls, so the tRPC client can handle it correctly. The two fixes are complementary.

PR #621, opened 2026-09-11
fix(s3): default empty region to us-east-1

bionicman:
## Description

While reviewing another PR, I noticed that the MinIO E2E setup uses a fixed `us-east-1` region. I checked whether self-hosted installations could override it and found that Kan already exposes `S3_REGION`, but its default still has a gap.

`S3_REGION` is optional and `.env.example` leaves it empty. The fallback added in #495 uses `??`, which handles an unset variable but not an empty string. Docker Compose can therefore pass `S3_REGION=""` to `S3Client`, which still throws `Region is missing`.

This normalizes an empty value before applying the existing `us-east-1` default. Explicitly configured regions remain unchanged. Regression tests cover unset, empty, and configured values.

## Type of change

- [x] Bug fix
- [ ] Refactor / chore

## Checklist

- [ ] I have linked the relevant approved issue (required for features)
- [x] I have followed the project’s code style

## Related issues

Follow-up to #495.

### DIFF
--- a/packages/shared/src/utils/s3.test.ts
+++ b/packages/shared/src/utils/s3.test.ts
@@ -1,0 +1,42 @@
+import { afterEach, describe, expect, it } from "vitest";
+
+import { createS3Client } from "./s3";
+
+const originalS3Region = process.env.S3_REGION;
+
+afterEach(() => {
+  if (originalS3Region === undefined) {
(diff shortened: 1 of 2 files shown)


hjball (approved):
lgtm

PR #623, opened 2026-09-13
fix: clear attachment upload spinner after successful upload

### LINKED ISSUE #559: Attachment upload spinner never clears after a successful upload
## Description

After successfully uploading a card attachment, the paperclip upload button keeps showing its loading spinner indefinitely. The file **has** actually been uploaded and appears in the attachment list, but the button itself stays in the "uploading" state until the page is refreshed (which remounts the component and resets state).

## Where

`apps/web/src/views/card/components/AttachmentUpload.tsx`, in `uploadFile`:

```tsx
const uploadFile = async (file: File) => {
  setUploading(true);

  try {
    ...
    await invalidateCard(utils, cardPublicId);
    showPopup({
      header: t`Attachment uploaded`,
      message: t`Your file has been uploaded successfully.`,
      icon: "success",
    });
    // setUploading(false) is missing here on the success path
  } catch {
    ...
    setUploading(false); // only reset on failure
  }
};
```

`setUploading(false)` is only called in the `catch` block. On success it's never reset, so the button spins forever until the component remounts (e.g. on page refresh).

Confirmed present in `v0.5.5`, `v0.6.0`, and `main`.

## Steps to reproduce

1. Open a card and click the paperclip icon to upload a file.
2. Select a file and let it upload successfully.
3. Observe the paperclip button keeps spinning (`isLoading`) even though the attachment now appears in the card.
4. Refresh the page — the spinner is gone (button resets to idle) and the attachment shows normally.

## Expected behavior

`setUploading(false)` should also run after a successful upload (e.g. in a `finally` block, or right after `showPopup` in the success path), so the button returns to its idle state immediately.

## Suggested fix

```tsx
const uploadFile = async (file: File) => {
  setUploading(true);

  try {
    ...
    await invalidateCard(utils, cardPublicId);
    showPopup({ ... });
  } catch {
    showPopup({ ... error ... });
  } finally {
    setUploading(false);
  }
};
```

anandghegde:
## Description

In `AttachmentUpload`, `setUploading(false)` was only called in the `catch` block. After a successful upload the paperclip button kept spinning, and the hidden file input stayed `disabled`, so you couldn't add another attachment until the card view remounted. This moves the reset into a `finally` so it runs on both paths.

I also added one assertion to the existing self-hosted attachments e2e spec: after the upload finishes, `#attachment-upload` should be enabled again.

Testing (locally, with Postgres, MinIO and Mailpit in Docker):
- Ran `tests/self-hosted/attachments.spec.ts` with the new assertion before the fix. It fails with the input still `disabled`.
- With the fix, the spec passes.
- `eslint` and `tsc --noEmit` pass for `@kan/e2e`. For `@kan/web`, the changed file lints clean. `tsc` shows the same errors with and without this change, and none of them are in files touched here.

Screenshots (card view about 2s after the "Attachment uploaded" toast):

| Before | After |
| --- | --- |
| Before | After |

## Type of change

- [x] Bug fix
- [ ] Refactor / chore
- [ ] Documentation

## Checklist

- [x] I have included screenshots for any UI changes

## Linked issue

Fixes #559


hjball (approved):
Nice one @anandghegde - will merge once e2es have finished

PR #629, opened 2026-09-14
Require board:create permission for GitHub project import (fixes #628)

### LINKED ISSUE #628: Privilege escalation: guest members can create boards via GitHub project import
## Summary

kan has three workspace roles (`admin`, `member`, `guest`) plus per-member permission overrides. `guest` is a by-design read-only role: its default permissions include `board:view` but not `board:create`. Board creation normally requires `board:create`, enforced via `assertPermission` in both direct board creation and the Trello import mutation (`importBoards`). The GitHub project-import mutation (`importProjects`), which also creates boards, instead checks only that the caller is any active member of the target workspace (`assertUserInWorkspace`, no role argument), not that they hold `board:create`. A `guest`, or a `member` with `board:create` individually revoked, can therefore create boards by importing GitHub projects while being correctly blocked on every other board-creation path.

Affected version: confirmed on commit `f08920d` (current `main`).

## Details

`packages/shared/src/permissions.ts`:

```ts
export const roleHierarchy = { admin: 100, member: 50, guest: 10 } as const;

export const defaultRolePermissions: Record<Role, readonly Permission[]> = {
  admin: allPermissions,
  member: [ ..., "board:view", "board:create", ... ],
  guest: [
    "workspace:view", "board:view", "list:view", "card:view",
    "comment:view", "member:view",
  ],
};
```

`packages/api/src/routers/import.ts`, Trello import (correctly gated):

```ts
await assertPermission(ctx.db, userId, workspace.id, "board:create");
```

GitHub import (missing the check):

```ts
await assertUserInWorkspace(ctx.db, userId, workspace.id);
```

`assertUserInWorkspace` (`packages/api/src/utils/auth.ts`) calls `isUserInWorkspace` with `role` undefined, which matches any active membership row regardless of role:

```ts
export const isUserInWorkspace = async (db, userId, workspaceId, role?) => {
  const result = await db.query.workspaceMembers.findFirst({
    where: and(
      eq(workspaceMembers.userId, userId),
      eq(workspaceMembers.workspaceId, workspaceId),
      eq(workspaceMembers.status, "active"),
      isNull(workspaceMembers.deletedAt),
      role ? eq(workspaceMembers.role, role) : undefined,
    ),
  });
  return result?.id !== undefined;
};
```

## POC

(available upon request)

## Impact

Any `guest` member, or any `member` with `board:create` explicitly revoked via a permission override, can create new boards in a workspace by importing GitHub projects (`POST /imports/github/projects`), bypassing the `board:create` authorization boundary enforced everywhere else. Unauthorized structural changes to a workspace by a user meant to have read-only access, and evasion of any admin-configured per-member permission restriction.

A fix PR (replacing the membership check with the same `assertPermission` check the Trello import already uses, plus a regression test) is attached.

carfeii:
Fixes #628.

## Problem

`importProjects` (GitHub project import, `packages/api/src/routers/import.ts`) creates a board per imported project but only checks `assertUserInWorkspace` (any active membership, no role), unlike the sibling Trello import (`importBoards`), which correctly requires `board:create` via `assertPermission`. Since `guest` is a real role whose default permissions exclude `board:create`, and `board:create` can also be individually revoked from a `member` via a permission override, a guest (or a restricted member) could create boards through the GitHub import path while being blocked everywhere else.

## Change

Replace the `assertUserInWorkspace` call in `importProjects` with the same `assertPermission(ctx.db, userId, workspace.id, "board:create")` call `importBoards` already uses, matching the existing pattern exactly. Drops the now-unused `assertUserInWorkspace` import.

## Tests

Adds `packages/api/src/routers/import.test.ts` (following the existing `webhook.test.ts` mocking pattern): confirms `importProjects` calls `assertPermission` with `board:create` and rejects when it's denied, and succeeds when it's granted.

Ran the full `packages/api` suite: 140/145 pass. The 5 failures (2 integration tests needing a live DB, 3 cross-test-isolation flakes) reproduce identically without this change and pass individually in isolation, confirmed pre-existing and unrelated. `tsc --noEmit` clean on the changed files.


hjball (approved):
Good catch @carfeii

